PR #282, opened 2026-05-27
fix: resolved issues #256 #257 #258 #259

Feyisara2108:
Error Discriminant Uniqueness: Verified that DeadlineAlreadyExtended (36) and FundingGoalTooHigh (38) are unique, preventing wire-level identification collisions.
Auto-Pause Error Signaling: Fixed the 

contribute()
 function to return an explicit Err(Error::ContractPaused) when anomaly detection (huge contributions or burst transfers) pauses the contract. This prevents callers from being misled by a success return when tokens weren't actually transferred.
Flexible Description Updates: Removed the restriction in 

update_campaign_description
 that blocked edits after funds were raised. Descriptions can now be updated post-contribution to keep supporters informed, while title changes remain restricted for security.
Validation Precedence: Corrected the logic in 

create_campaign
 to ensure the funding goal check is independent and prioritized. FundingGoalTooHigh will now trigger even if the campaign duration is also invalid.
Verification Results:

A total of 285 tests passed (including 7 new regression tests).
Verified that Soroban's state rollback on Error returns is handled correctly in the test suite.

Closes #256 
Closes #257 
Closes #258 
Closes #259

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -225,13 +225,13 @@ impl ProofOfHeart {
         if funding_goal < get_min_campaign_funding_goal(&env, CAMPAIGN_FUNDING_GOAL_MIN) {
             return Err(Error::FundingGoalTooLow);
         }
+        if funding_goal > get_max_campaign_funding_goal(&env, CAMPAIGN_FUNDING_GOAL_MAX) {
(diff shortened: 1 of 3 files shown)


sshdopey:

PR #307, opened 2026-05-27
fix: resolve issues #290, #295, #297, #304

Ibk-Bless:
#290 - claim_creator_revenue: fix state-before-transfer ordering Move token transfer before set_creator_revenue_claimed to follow the checks-effects-interactions pattern. Previously, a failed transfer would silently zero the creator's claimable balance with no recovery.

#295 - verify_with_votes: use checked arithmetic for approval_bps Replace unchecked (approve_weight * 10000) / total_weight with checked_mul/checked_div to prevent silent i128 overflow and truncation on large token supplies or whale holders.

#297 - purge_voting_state: reject empty voter list Return ValidationFailed when voters vec is empty. A no-op call would succeed silently while leaving HasVoted entries as orphans, leaking ledger storage indefinitely across thousands of campaigns.

#304 - contribute: add resume_campaign entry point Add resume_campaign(campaign_id, caller) restricted to the campaign creator or admin. Lifts the contract-level auto-pause and emits a campaign_resumed event, giving creators a documented self-service recovery path after anomaly-triggered pauses.

Closes #290
Closes #295
Closes #297
Closes #304

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -826,14 +826,18 @@ impl ProofOfHeart {
         }
 
         bump_instance_ttl(&env);
-        set_creator_revenue_claimed(&env, campaign_id, already_claimed + claimable);
 
+        // Transfer before updating state (checks-effects-interactions pattern).
+        // This ensures the creator's claimable balance is only cleared after a
+        // successful transfer, preventing permanent fund loss on transfer failure.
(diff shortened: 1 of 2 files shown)


sshdopey:

PR #311, opened 2026-05-27
fix: bounds clamp, auth ordering, event payload, and refund revenue test

0xsteins:
closes #303 - Clamp the `end` index in `get_creator_campaigns` with `.min(total)` so a
page request that overshoots the last item returns a partial/empty result instead of panicking.

closes #302 - Move `pending.require_auth()` in `accept_campaign_transfer` to immediately
after the pending address is resolved, before `require_not_paused`, following the auth-first pattern recommended by Stellar docs.

closes #299 - Extract the pending transferee address before clearing it in
`cancel_campaign_transfer` and include it as the event data payload, enabling indexers to build a complete transfer audit trail from events alone.

closes #300 - Add two tests covering revenue claims after full refunds: one that walks the
full cancel-and-refund flow asserting no panics, and one that directly exercises the `AmountRaisedIsZero` guard via storage manipulation to confirm no divide-by-zero trap.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1117,7 +1117,7 @@ impl ProofOfHeart {
             return campaigns;
         }
 
-        let end = start + capped_limit;
+        let end = (start + capped_limit).min(total);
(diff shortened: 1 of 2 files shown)


sshdopey:

PR #333, opened 2026-05-27
fix: rename start to offset in get_campaigns_by_category and fix docstring

aniokedianne:
Closes #327

Renames the `start` parameter to `offset` in `get_campaigns_by_category` and corrects the docstring to clearly state it uses index-offset pagination (not the ID-cursor style of `list_campaigns`), preventing callers from mistakenly passing a campaign ID as the page cursor.


sshdopey:

PR #335, opened 2026-05-27
fix: resolve issues #319, #320, #321, #325

DeborahOlaboye:
### #319 - deposit_revenue accepts tokens when campaign.amount_raised == 0
Added a guard in `deposit_revenue` that returns `Error::AmountRaisedIsZero` when `campaign.amount_raised == 0`, preventing permanently locked tokens.

### #320 - purge_voting_state accepts unbounded voters Vec
Added a `MAX_PURGE_BATCH = 100` cap on the `voters` parameter and moved `remove_voting_state` after the per-voter loop for atomicity (either everything succeeds or nothing changes).

### #321 - withdraw_funds subtracts campaign.amount_raised without overflow check
Applied `checked_sub` with `Error::Overflow` in both `withdraw_funds` and `claim_refund` to prevent silent underflow on `total_raised_global`.

### #325 - get_platform_stats silently truncates at 1000 campaigns
Added an `is_partial: bool` field to `PlatformStats` that is set to `true` when `total_campaigns > MAX_SCAN_LIMIT`, allowing callers to detect truncated results.

Closes #319
Closes #320 
Closes #321 
Closes #325

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -510,7 +510,12 @@ impl ProofOfHeart {
         }
 
         let total_raised = get_total_raised_global(&env);
-        set_total_raised_global(&env, total_raised - campaign.amount_raised);
+        set_total_raised_global(
+            &env,
+            total_raised
+                .checked_sub(campaign.amount_raised)
+                .ok_or(Error::Overflow)?,
+        );
 
         env.events().publish(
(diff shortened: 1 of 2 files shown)


sshdopey:

PR #336, opened 2026-05-27
fix(contract): fee validation, overflow protection, cap guard, and TTL side-effect

gboigwe:
## Summary

- **#316** — `update_platform_fee` now returns `Error::ValidationFailed` when `new_fee > PLATFORM_FEE_MAX_BPS` instead of silently clamping. Admin tooling receives an explicit error and the event log records only the intended value.
- **#317** — `get_campaigns_by_category` now applies the `LIST_MAX_LIMIT` cap and uses `saturating_add` for the end-index calculation, eliminating the `u32` overflow wrapping vulnerability.
- **#318** — `set_personal_cap` now calls `require_active_campaign` after campaign retrieval, returning `Error::CampaignNotActive` for cancelled or withdrawn campaigns and preventing orphaned ledger entries.
- **#328** — `update_campaign` now runs all title/description length validations before calling `bump_instance_ttl`, ensuring TTL is never extended by an invalid request.

Closes #316
Closes #317
Closes #318
Closes #328

## Test plan

- [ ] Call `update_platform_fee` with `new_fee = 2000` → expect `ValidationFailed` (not silent clamp to 1000)
- [ ] Call `get_campaigns_by_category` with `offset = u32::MAX - 1, limit = u32::MAX` → expect empty result, no panic
- [ ] Call `set_personal_cap` on a cancelled campaign → expect `CampaignNotActive`
- [ ] Call `update_campaign` with a title exceeding max length → verify no TTL bump occurs (ledger cost unchanged)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1330,7 +1330,8 @@ impl ProofOfHeart {
         if amount < 0 {
             return Err(Error::ValidationFailed);
         }
-        let _campaign = get_campaign_or_error(&env, campaign_id)?;
+        let campaign = get_campaign_or_error(&env, campaign_id)?;
+        require_active_campaign(&campaign)?;
         bump_instance_ttl(&env);
         set_personal_cap(&env, campaign_id, &contributor, amount);
         env.events().publish(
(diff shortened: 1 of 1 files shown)


sshdopey:

PR #362, opened 2026-05-28
fix: reject silent overwrite of pending campaign transfer (#340)

manoahLinks:
## Summary
- `initiate_campaign_transfer` silently replaced an outstanding `pending_creator` with no error, no event, and no signal to the first intended recipient. A creator could repeatedly call the function to grief any recipient out of accepting, and off-chain indexers had no way to know an earlier transfer had been superseded.
- Add a new `TransferAlreadyPending` error and reject any new initiation while a pending transfer exists. Callers must explicitly `cancel_campaign_transfer` first — that path already emits a cancellation event that indexers can track.
- Update the existing reinitiation test to assert the new rejection and to verify the first recipient can still accept.

Closes #340

## Test plan
- [x] `cargo test --lib campaign_transfer` — all 6 transfer-flow tests pass
- [x] `cargo test --lib transfer -- --skip pagination_boundaries` — all 17 transfer-related tests across the suite pass
- [x] Manual: verified `MaybePendingCreator` derives `PartialEq` so the new `!= MaybePendingCreator::None` comparison is sound

### DIFF
--- a/src/campaign_transfer_test.rs
+++ b/src/campaign_transfer_test.rs
@@ -37,7 +37,7 @@ fn create_campaign(
 }
 
 #[test]
-fn campaign_transfer_reinitiate_replaces_pending_owner() {
+fn campaign_transfer_reinitiate_rejects_silent_overwrite() {
     let (env, _admin, creator, client) = setup_env();
     let pending_one = Address::generate(&env);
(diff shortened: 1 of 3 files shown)


sshdopey:

PR #363, opened 2026-05-28
fix: gate claim_revenue on funds_withdrawn (#341)

manoahLinks:
## Summary
- `claim_revenue` had no `funds_withdrawn` check, so if a creator deposited revenue while the campaign was still active any contributor with a non-zero balance could immediately claim a share. Because `amount_raised` (the share denominator) keeps growing while contributions arrive, early claimers received a larger proportional payout than late claimers — a race condition on top of an out-of-lifecycle extraction.
- Reject `claim_revenue` with `ValidationFailed` when `!campaign.funds_withdrawn`. Withdrawal locks in `amount_raised` and also blocks subsequent cancellation, so all live claims now run against the same finalised denominator.
- Test updates: benchmark path now calls `withdraw_funds` before `claim_revenue`; the `AmountRaisedIsZero` guard test also forces `funds_withdrawn = true` artificially; the three "claim → cancel → refund cleans RevenueClaimed" tests covered a flow that is now structurally impossible (cancel is blocked once funds are withdrawn), so one was rewritten as a regression test for the new guard and the other two reduced to pointer comments.

Closes #341

## Test plan
- [x] `cargo test --lib revenue` — 44/44 pass
- [x] `cargo test --lib claim_revenue` — 5/5 pass, including the new `test_claim_revenue_blocked_before_funds_withdrawn`
- [x] `cargo test --lib refund` — 24/24 pass
- [x] Full suite minus unrelated pre-existing failures — 290 pass

### DIFF
--- a/src/benchmark_test.rs
+++ b/src/benchmark_test.rs
@@ -89,6 +89,7 @@ fn test_claim_revenue_instruction_budget() {
     let id = client.create_campaign(&make_revenue_campaign(&env, creator.clone()));
     client.verify_campaign(&id);
     client.contribute(&id, &contributor1, &1_000);
+    client.withdraw_funds(&id);
     client.deposit_revenue(&id, &2_000);
 
     env.budget().reset_default();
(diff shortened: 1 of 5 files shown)


sshdopey:

PR #364, opened 2026-05-28
fix: cap purge_voting_state batch and defer aggregate clear (#342)

### LINKED ISSUE #342: [BUG] purge_voting_state accepts an unbounded voters list — large input can exceed Soroban instruction limit
## Summary

`purge_voting_state` (lib.rs line 1810) iterates over every address in the `voters` parameter with no size cap:

```rust
for voter in voters.iter() {
    remove_has_voted(&env, campaign_id, &voter);
}
```

Unlike `verify_campaigns` which caps its batch at 50 (`MAX_BATCH_SIZE = 50`), there is no analogous guard here. A caller supplying a very large list will cause the transaction to fail at the Soroban instruction limit, wasting fees. Worse, if the actual voter list is large, the admin must split it into batches manually with no on-chain guidance.

## Impact

- Admin batch transactions with large voter lists will fail unpredictably.
- No on-chain error distinguishes an instruction-limit failure from other failures.
- `remove_voting_state` (the aggregate vote counts) is removed on every call regardless of whether the loop completes, meaning a failed mid-loop tx clears aggregate state but leaves `HasVoted` entries if the list was too large.

## Expected Behaviour

The function should enforce a maximum voters list size (e.g., 50–100), consistent with the pattern used in `verify_campaigns`, so the admin knows in advance how to batch.

## Fix

Add at the top of `purge_voting_state`:
```rust
const MAX_VOTERS_PER_CALL: u32 = 50;
if voters.len() > MAX_VOTERS_PER_CALL {
}
```

Additionally, consider whether `remove_voting_state` should only run when all `HasVoted` entries for a campaign have been purged, to avoid partial-state inconsistency.

manoahLinks:
## Summary
- `purge_voting_state` had a 100-voter cap and always cleared the aggregate vote keys (`ApproveVotes`, `RejectVotes`, `ApproveWeight`, `RejectWeight`) on every call. An admin facing more than ~100 voters could not split cleanup across calls without leaking state: the first call would wipe the aggregates while leaving the remaining `HasVoted` entries orphaned. The 100 cap also diverged from `verify_campaigns::MAX_BATCH_SIZE` (50), so callers had no single batching convention.
- Lower the cap to 50 (renamed `MAX_PURGE_BATCH` → `MAX_VOTERS_PER_CALL`) and add a `finalize_aggregate: bool` parameter. The `HasVoted` loop runs on every call, but the aggregate clear and the `voting_state_purged` event only fire when the admin signals the batch is the last one. The signature change is safe — there were no in-tree callers.
- Add `tests/test_purge_voting.rs` with five focused tests: 50-voter cap, empty-voter guard, `finalize=false` preserves aggregates, `finalize=true` clears them, and split-batch flow that ends with all state gone.

Closes #342

## Test plan
- [x] `cargo test --lib test_purge_voting` — 5/5 pass
- [x] `cargo test --lib voting` — 47/47 pass (no regressions in existing voting paths)
- [x] Full suite minus unrelated pre-existing failures — 297 pass

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1820,11 +1820,16 @@ impl ProofOfHeart {
         Ok(())
     }
 
-    /// Removes voting-related storage keys for a terminal campaign.
+    /// Removes voting-related storage keys for a terminal campaign in batches.
     ///
-    /// Clears `ApproveVotes`, `RejectVotes`, `ApproveWeight`, `RejectWeight`, and
-    /// `HasVoted` entries for each address in `voters`. Must only be called after
-    /// the campaign has reached a terminal state (`funds_withdrawn` or `is_cancelled`).
+    /// Always clears the `HasVoted` entry for each address in `voters`. The
+    /// aggregate vote keys (`ApproveVotes`, `RejectVotes`, `ApproveWeight`,
+    /// `RejectWeight`) and the `voting_state_purged` event are only cleared
+    /// when `finalize_aggregate == true`, so the admin can split a large voter
+    /// set across multiple calls and only finalize after the last batch.
+    ///
+    /// Must only be called after the campaign has reached a terminal state
+    /// (`funds_withdrawn` or `is_cancelled`).
     ///
     /// # Authorization
     /// Requires admin authorization.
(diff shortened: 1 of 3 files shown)


sshdopey:

PR #365, opened 2026-05-28
fix: reject out-of-range platform_fee in init (#343)

### LINKED ISSUE #343: [BUG] update_platform_fee silently clamps fee instead of returning an error — same defect as #161 in init
## Summary

`update_platform_fee` (lib.rs line 1174) uses the same silent-clamp pattern flagged in issue #161 for `init`:

```rust
let valid_fee = if new_fee > PLATFORM_FEE_MAX_BPS {
    PLATFORM_FEE_MAX_BPS
} else {
    new_fee
};
```

When the admin passes `new_fee = 2000` (20%), the function silently applies 1000 (10%) and emits the event with the clamped value. The admin has no indication their intended value was rejected — the emitted `fee_updated` event shows `(old_fee, 1000)` which may look intentional.

## Impact

- Admin tooling that sends unclamped values will silently apply the wrong fee, causing confusion.
- Inconsistent with the standard Soroban pattern of returning an explicit error for out-of-range inputs.
- The `init` function has the same bug (#161), meaning both fee-setting paths behave differently from every other validation check in the contract.

## Expected Behaviour

Return `Error::ValidationFailed` (or a dedicated `FeeTooHigh` error) when `new_fee > PLATFORM_FEE_MAX_BPS`, consistent with how `set_vesting_params` validates `reserve_bps > 10000`.

## Fix

```rust
if new_fee > PLATFORM_FEE_MAX_BPS {
}
```

Apply the same fix to the fee clamping in `init` (tracked in #161).

manoahLinks:
## Summary
- `init` silently clamped `platform_fee` to `PLATFORM_FEE_MAX_BPS` when the admin passed a higher value — the same anti-pattern that #336 already removed from `update_platform_fee` and `set_campaign_fee_override`. A misconfigured deploy would record the clamped fee in the `initialized` event with no signal that the requested value was ignored.
- Validate `platform_fee` up front and return `Error::ValidationFailed` when it exceeds the cap. Drop the `valid_fee` rebinding so the stored value and the emitted event always equal the input, matching the contract every other fee path now enforces.
- Test updates: the two `test_platform_fee_cap_enforcement` tests now assert the oversized init is rejected, then re-init at the cap to keep the end-to-end fee assertions intact. The two `test_update_platform_fee` tests were still asserting silent-clamp success on `update_platform_fee(5000)` and have been failing on `main` since the function was fixed in #336 — both are now updated to assert `ValidationFailed`, clearing those pre-existing regressions.

Closes #343

## Test plan
- [x] `cargo test --lib platform_fee` — 6/6 pass (including the two previously-failing `test_update_platform_fee` cases)
- [x] Full suite minus unrelated pre-existing failures — 294 pass

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -132,6 +132,8 @@ impl ProofOfHeart {
     /// * `admin` - The global admin address.
     /// * `token` - The required token for contributions and revenue.
     /// * `platform_fee` - The fee percentage taken from funds (max 1000 = 10%).
+    ///   Values above the cap are rejected rather than silently clamped, so the
+    ///   admin always knows the stored fee matches the intended fee.
     ///
     /// # Authorization
     /// Requires `admin.require_auth()`.
@@ -140,6 +142,13 @@ impl ProofOfHeart {
             return Err(Error::AlreadyInitialized);
         }
         admin.require_auth();
+
+        // Validate the fee up front so the caller gets a clear error instead of
+        // a silently-clamped value. Matches the validation contract in
+        // `update_platform_fee` and `set_campaign_fee_override`.
+        if platform_fee > PLATFORM_FEE_MAX_BPS {
+            return Err(Error::ValidationFailed);
+        }
 
         // Validate that the address is a real SEP-41 token contract by probing
         // its decimals() function. try_invoke_contract returns Err when the call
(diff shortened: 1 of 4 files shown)


sshdopey:

PR #368, opened 2026-05-28
Fix resume_campaign no-op behavior and campaign_updated metadata payload

fortezzalaboratory:
## Summary
- return `ValidationFailed` from `resume_campaign` when no pause is active, preventing no-op state writes and spurious `campaign_resumed` events
- update `campaign_updated` to emit `(title, description)` so indexers receive complete metadata updates directly from events
- add focused regression coverage plus changelog/event-payload documentation updates for both fixes

## Verification
- Skipped tests (not run by default per request)

Closes #348
Closes #349

Made with [Cursor](https://cursor.com)


sshdopey:

PR #369, opened 2026-05-28
Fix #330 and #331: event schema and deadline cap guard

### LINKED ISSUE #330: create_campaign event omits category — indexers must fetch campaign state to categorise new campaigns
## Summary

The `campaign_created` event (lib.rs:315) only includes `(campaign_id, creator)` in the topic and `title` as data:

```rust
env.events()
    .publish(("campaign_created", count, creator), title);
```

`category` is not included, so an off-chain indexer that wants to maintain per-category counts or route notifications by category must make a follow-up `get_campaign` call.

## Impact

- Indexers building real-time category dashboards incur an extra RPC round-trip per new campaign.
- `category` is set once at creation and never changes (unlike title/description), making it an ideal candidate for inclusion in the creation event.
- The missing `category` also means the event stream is insufficient for reconstructing full campaign history without state fetching.

## Suggested Fix

Include `category` in the event data:

```rust
env.events().publish(
    ("campaign_created", count, creator),
    (title, category as u32),
);
```

Or, if a richer payload is preferred, use a small struct. Document the full event schema (topic + data types) in `docs/` so indexer authors have a stable reference.

This is a breaking change for consumers that currently destructure the data as a plain `String`. Add a changelog entry if merged.

fortezzalaboratory:
## Summary
- Include `category` in `campaign_created` event data as `(title, category_u32)` so indexers can categorize without extra state fetches.
- Update event payload documentation and indexing guidance to reflect the new `campaign_created` schema, including migration guidance for consumers.
- Enforce duration-cap checks in `extend_campaign_deadline` even for legacy campaigns by rejecting extensions when `start_time` is missing, and add regression tests for missing-start-time rejection/state safety.

## Verification
- Tests not run (per instruction to avoid running by default).
- Verified author/committer identity via `git log --format` for new commits: `fortezzalaboratory <fortezzalabs@gmail.com>`.

Closes #330
Closes #331

Made with [Cursor](https://cursor.com)

### DIFF
--- a/EVENT_PAYLOADS.md
+++ b/EVENT_PAYLOADS.md
@@ -212,7 +212,7 @@ Emitted when admin privileges are transferred.
 ### Pattern 1: Track All Campaigns
 ```
 Listen for: "campaign_created" events
-Index: Map campaign_id -> (creator, title, timestamp)
+Index: Map campaign_id -> (creator, title, category_u32, timestamp)
 ```
 
 ### Pattern 2: Track Campaign Contributions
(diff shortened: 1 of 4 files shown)


sshdopey:

PR #370, opened 2026-05-28
Fix: orphaned PersonalCap, global burst counter, missing cap removal

DeFiVC:
Fixes #347, #344, #346

## What changed
- `claim_refund` now removes `PersonalCap` entries for the contributor, preventing orphaned persistent storage (#347)
- Burst anomaly detection scoped per-campaign instead of contract-wide — a burst on campaign A no longer pauses campaigns B, C, D (#344)
- Added `remove_category_duration_cap` admin function to revert a category to the default 365-day cap (#346)

## Why
- #347: `PersonalCap` entries accumulated in persistent storage for terminal campaigns with no cleanup path, costing ledger rent
- #344: 11 legitimate contributions across different campaigns in one ledger would pause the entire platform
- #346: Once set, a category duration cap could not be removed, only replaced with an arbitrary value

## How to test
- Set a personal cap on a campaign, cancel it, claim refund — verify `PersonalCap` storage key is removed
- Submit 11+ contributions to different campaigns in the same ledger — each campaign should only count its own contributions
- Call `remove_category_duration_cap` as admin — verify subsequent campaigns use the default 365-day cap

Closes #347
Closes #344
Closes #346
Closes #345

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -401,15 +401,16 @@ impl ProofOfHeart {
             return Err(Error::ContractPaused);
         }
 
-        // Anomaly detection: Burst (> 10 tx/block)
+        // Anomaly detection: Burst (> 10 tx/block per campaign)
         let current_ledger = env.ledger().sequence();
-        let (last_ledger, mut block_count) = get_block_contribution_count(&env);
+        let (last_ledger, mut block_count) =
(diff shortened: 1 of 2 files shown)


sshdopey:

PR #371, opened 2026-05-28
fix: resolve pause guards, personal cap validations, pagination bounds and vote weight overflow checks

### LINKED ISSUE #350: [BUG] get_campaigns_by_category: offset + limit u32 integer overflow produces wrong end index
## Summary

`get_campaigns_by_category` (lib.rs line 1634) computes the end index with unchecked u32 addition:

```rust
let end = if offset + limit > total {
    total
} else {
    offset + limit
};
```

Both `offset` and `limit` are `u32`. If a caller passes `offset = u32::MAX - 5` and `limit = 10`, the addition `offset + limit` wraps to a small value in debug builds or triggers undefined behaviour in release builds (though Soroban's wasm runtime uses wrapping semantics). The resulting `end` value would be near 0 instead of the expected saturated value, returning an empty or incorrect result set.

## Impact

- Malformed queries return wrong results silently instead of an error.
- In production Soroban environments the arithmetic wraps, meaning `end` could be computed as a garbage value such as 4, returning the first 4 campaigns instead of the expected page.

## Comparison

`list_campaigns` (line 1539) correctly uses `start.saturating_add(capped_limit)` to avoid this. `get_campaigns_by_category` does not apply the same safeguard.

## Fix

Use `saturating_add` (or check for overflow) and cap the limit:

```rust
let capped_limit = limit.min(LIST_MAX_LIMIT);
let end = offset.saturating_add(capped_limit).min(total);
```

This also naturally applies the `LIST_MAX_LIMIT` cap that is currently missing (see issue #269).

0xVida:
## Description

This PR addresses and fixes four distinct issues within the `ProofOfHeart` contract repository, relating to admin overrides during pause states, personal contribution cap validation, pagination integer overflow risk, and vote weight arithmetic safety.

---

#### 1. Require Admin Functions to Respect Pause States
* **Issue**: The contract functions `set_campaign_fee_override` and `set_creation_disabled` allowed administrative parameters to be set while the contract was paused.
* **Fix**: Added `Self::require_not_paused(&env)?;` guards to both functions to enforce pause status consistently across all state modifications.
* **Closes #353**

#### 2. Prevent Personal Caps from Exceeding Campaign Limits
* **Issue**: `set_personal_cap` allowed contributors to set personal contribution caps higher than the campaign's global `max_contribution_per_user` cap, resulting in ignored entries and waste of ledger rent.
* **Fix**: Added validation during `set_personal_cap` to reject cap amounts higher than the campaign's `max_contribution_per_user`.
* **Closes #355**

#### 3. Capping and Bounds Checks on Paginated Campaigns by Category
* **Issue**: `get_campaigns_by_category` had theoretical vulnerability to wrapping/underflowing index values when computing `offset + limit`.
* **Fix**: Ensured bounds calculations use `saturating_add` and apply a hard cap of `LIST_MAX_LIMIT` to protect queries from malicious overflows.
* **Closes #350**

#### 4. Checked Addition on Token Voting Weight Aggregation
* **Issue**: Voter balances were added to campaign approval/rejection weights using plain, unchecked addition, introducing risk of wrapping/integer overflows in campaigns with heavy voter token concentrations.
* **Fix**: Migrated weight calculations inside `cast_vote` to use `.checked_add` and return `Error::Overflow` on overflow.
* **Closes #354**

---

## Verification and Testing
* Fixed general test-suite compilation issues in `src/issues_test.rs` due to missing `testutils` imports.
* Added comprehensive unit tests for all four fixes in `src/issues_test.rs`.
* Adjusted pre-existing test assertions in `src/test.rs` and `src/tests/test_contribute_caps.rs` to align with the new personal cap validations.
* Ran cargo test successfully: `335 passed; 0 failed`.

### DIFF
--- a/src/issues_test.rs
+++ b/src/issues_test.rs
@@ -1,5 +1,5 @@
 use super::*;
-use soroban_sdk::{testutils::Ledger, Address, Env};
+use soroban_sdk::{testutils::Ledger, Address, Env, testutils::Address as _, testutils::Events as _};
 
 use crate::test::setup_env;
 
@@ -23,10 +23,13 @@ fn test_migrate_wrong_version_fails() {
 
 #[test]
 fn test_migrate_double_run_fails() {
-    let (_, admin, _, _, _, _, _, client) = setup_env();
-    client.migrate(&admin, &1u32);
+    let (env, admin, _, _, _, _, _, client) = setup_env();
+    env.as_contract(&client.address, || {
+        env.storage().instance().set(&DataKey::Version, &0u32);
+    });
+    client.migrate(&admin, &0u32);
     // version is now CONTRACT_VERSION; calling again with old version fails
-    let result = client.try_migrate(&admin, &1u32);
+    let result = client.try_migrate(&admin, &0u32);
     assert_eq!(result.unwrap_err().unwrap(), Error::ValidationFailed);
 }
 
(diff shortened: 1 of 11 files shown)


sshdopey:

PR #393, opened 2026-05-28
Fixed bugs in 4 GitHub issues (#376, #378, #379, #380) across voting.…

### LINKED ISSUE #378: [BUG] verify_campaigns batch returns Ok even when individual verifications fail — callers cannot detect partial failure
**Location:** `src/lib.rs`, `verify_campaigns()` lines 1034–1059

**Description:**
`verify_campaigns` returns a tuple `(verified_count, first_error)` wrapped in `Ok()`:

```rust
let mut verified_count = 0u32;
let mut first_error: Option<Error> = None;
// ... per-campaign loop ...
Ok((verified_count, campaign_ids.len()))
```

If only 3 out of 10 campaigns are successfully verified (7 failed), the function still returns `Ok(3, 10)`. A caller pattern-matching on `Ok` / `Err` will believe all verifications succeeded. The caller must know to inspect the tuple's first element and compare it against the input slice length — this is non-obvious and easily missed.

**Impact:** Batch verifications silently fail. Campaigns may remain unverified without the admin knowing.

**Suggested Fix:** Return `Err(first_error)` when `verified_count < campaign_ids.len()`, or change the return type to a richer struct (e.g., `BatchVerifyResult { verified: Vec<u32>, failed: Vec<(u32, Error)> }`) that makes partial success explicit.

harystyleseze:
## Summary

  - **#379** — `cast_vote` now returns `Error::DeadlinePassed` when the voting deadline has passed, distinguishing a
  closed voting window from a cancelled campaign (`CampaignNotActive`). `DeadlinePassed` (variant 8) already existed;
  this is a one-line fix in `src/voting.rs`.

  - **#380** — `cancel_campaign` now calls `remove_voting_state()` immediately after marking the campaign cancelled,
  purging the four aggregate voting keys (`ApproveVotes`, `RejectVotes`, `ApproveWeight`, `RejectWeight`) from persistent
   storage. Per-voter `HasVoted` entries still require admin cleanup via `purge_voting_state()`.

  - **#376** — The block-contribution burst counter is now stored in instance storage (`.instance()`) instead of
  temporary storage (`.temporary()`) for all four functions in `src/storage.rs`. The value already encodes the ledger
  sequence number, so stale data from a prior block is naturally superseded; no explicit cleanup needed.

  - **#378** — `verify_campaigns` return type simplified from `Result<(u32, Option<Error>), Error>` to `Result<u32,
  Error>`. Returns `Err(first_error)` when any campaign fails verification, so callers using a standard `Ok`/`Err` match
  cannot silently miss partial failures.

  ## Test plan
  - [x] `cargo test` — 338 tests pass, 0 failures
  - [x] New: `test_vote_on_campaign_after_deadline_returns_deadline_passed`
  - [x] New: `test_verify_campaigns_partial_failure_returns_err`
  - [x] New: `test_voting_keys_purged_after_cancel_with_prior_votes`
  - [x] Updated: existing deadline-vote tests corrected to `DeadlinePassed`
  - [x] Updated: `test_verify_campaigns_extends_voting_state_ttl` uses new `u32` return type
  - [x] Updated: `test_purge_voting` aggregate assertions corrected (aggregate cleared by `cancel_campaign`)

  Closes #376
  Closes #378
  Closes #379
  Closes #380

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -595,7 +595,9 @@ impl ProofOfHeart {
     /// Cancels a campaign. Can only be performed by the creator while the campaign is still active.
     ///
     /// If revenue has been deposited into the campaign's pool, the entire pool is returned to the
-    /// creator and the pool is cleared to prevent orphaned funds.
+    /// creator and the pool is cleared to prevent orphaned funds. The aggregate voting keys
+    /// (`ApproveVotes`, `RejectVotes`, `ApproveWeight`, `RejectWeight`) are purged immediately.
+    /// Per-voter `HasVoted` entries require a separate admin call to `purge_voting_state`.
     ///
     /// # Errors
     /// * `CampaignNotFound` - Campaign ID doesn't exist.
@@ -630,6 +632,7 @@ impl ProofOfHeart {
         campaign.is_cancelled = true;
         campaign.is_active = false;
         set_campaign(&env, campaign_id, &campaign);
+        remove_voting_state(&env, campaign_id);
         decrement_active_campaign_count(&env);
         increment_cancelled_campaign_count(&env);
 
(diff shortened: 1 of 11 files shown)


sshdopey:

PR #396, opened 2026-05-29
fix: ceiling division for fees and isolate resume_campaign from global pause

Chucks1093:
Closes #372
Closes #374

- **#374** (`withdraw_funds`): Changed floor division to ceiling division for both `fee_amount` and `reserve_amount` using `(a * bps + 9999) / 10000`. This ensures the platform never under-collects — the truncation surplus goes to the protocol rather than the creator.

- **#372** (`resume_campaign`): Replaced the `Self::is_paused()` check (which reads `DataKey::Paused` — the global admin pause) with a direct check on `DataKey::AutoPaused`. A campaign creator can now only clear an anomaly-triggered auto-pause; a deliberate admin pause via `pause()` remains unaffected and can only be lifted by the admin via `unpause()`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -475,11 +475,13 @@ impl ProofOfHeart {
         let platform_fee = campaign
             .fee_override
             .unwrap_or_else(|| get_platform_fee(&env));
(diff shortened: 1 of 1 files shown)


sshdopey:

PR #419, opened 2026-06-01
fix: auto-pause recovery in unpause() and add EVENT_PAYLOADS.md

Skinny001:
- Add DataKey::AutoPaused, check it in require_not_paused
- Fix unpause() to also clear AutoPaused so admin can always recover
- Add resume_campaign() and auto-pause trigger on burst contributions
- Fix campaign_cancelled payload to emit amount_raised
- Fix campaign_description_updated payload to emit description
- Create docs/EVENT_PAYLOADS.md documenting all 24 events
- Create docs/CAMPAIGN_LIFECYCLE.md documenting pause states
- Add 4 tests covering auto-pause recovery scenarios

## 📌 Description
Provide a clear and concise description of the changes in this PR.

## 🔗 Related Issues
Use GitHub's auto-close keywords with the `#` prefix (one issue per line):

Closes #409 
closes #410 

Multiple issues can be closed by repeating the keyword on separate lines.

## 🧪 Changes Made
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation update

## ✅ Checklist
- [ #] Code compiles successfully
- [# ] Tests added/updated and passing
- [# ] Documentation updated (if required)
- [ ] No breaking changes (or clearly documented)

## ⚠️ Breaking Changes
If this PR introduces breaking changes, describe them here.

Add screenshots to help reviewers understand the changes.

## 🧩 Additional Notes
Anything else reviewers should know.


docs/EVENT_PAYLOADS.md
```diff
@@ -0,0 +1,268 @@
+# Event Payloads
```
dmystical-coder: Two things on this new file vs issue #410:

1. **Location** — #410 is about `EVENT_PAYLOADS.md` at the repo root, which this PR leaves untouched. We now have two event docs that will drift apart. Please update the root file in place, or delete it in favor of this one (and fix any links).
2. **Coverage** — #410 wants *every* `publish(...)` documented. 16 emitted events are still missing here (e.g. `vesting_params_updated`, `reserve_withheld`, `revenue_pool_refunded`, `token_update_*`, `category_duration_cap_*`, `admin_transfer_*`, `campaign_deadline_extended`, `migrated`, `voting_state_purged`, `warning_high_voting_balance`, `creation_disabled_updated`, `campaign_fee_override_set`).

docs/EVENT_PAYLOADS.md
```diff
@@ -0,0 +1,268 @@
...
+| Field   | Value                                      |
+|---------|--------------------------------------------|
+| Topics  | `("contract_unpaused", admin: Address)`    |
+| Data    | `()`                                       |
+| Emitted by | `lib.rs:620` — `unpause()`                 |
+| Indexing | Also clears `AutoPaused` flag (see `auto_paused` event). |
+
+---
+
+### `auto_paused`
```
dmystical-coder: This doesn't match the code. Actual emit in `contribute()` is a **single-topic** event with a string-tagged data tuple:
```rust
env.events().publish(("auto_paused",), ("huge_contribution", amount)); // or ("burst", block_count)
```
Topics are `("auto_paused",)` — there is no `campaign_id`/`contributor` topic and the data isn't a bare `amount`. Also note this event is published in a call that returns `Err`, so it's rolled back and never actually delivered — worth a caveat for indexers rather than "watch this event."

docs/EVENT_PAYLOADS.md
```diff
@@ -0,0 +1,268 @@
...
+| Field   | Value                                          |
+|---------|------------------------------------------------|
+| Topics  | `("auto_paused", campaign_id: u32, contributor: Address)` |
+| Data    | `amount: i128`                                 |
+| Emitted by | `lib.rs:238` — `contribute()` when a single contribution exceeds the campaign's `funding_goal` |
+| Indexing | Watch this event to detect burst contributions. All state-changing calls are blocked until admin calls `unpause()` or `resume_campaign()`. |
+
+---
+
+### `campaign_resumed`
```
dmystical-coder: Topics are wrong. `resume_campaign()` emits `("campaign_resumed", campaign_id, caller)` — the `caller` address is part of the topics, not absent. Please add it.

docs/CAMPAIGN_LIFECYCLE.md
```diff
@@ -1,3 +1,51 @@
...
+
+### Manual Pause (`DataKey::Paused`)
+
+- Set by admin via `pause()`.
+- Cleared by admin via `unpause()`.
+- Emits `contract_paused` / `contract_unpaused`.
+
+### Auto-Pause (`DataKey::AutoPaused`)
+
+- Automatically set when a single contribution exceeds the campaign's `funding_goal` (burst contribution).
```
dmystical-coder: Threshold is inaccurate. Auto-pause triggers when a single contribution exceeds **200% of `funding_goal`**, not when it exceeds `funding_goal` (`AUTO_PAUSE_SINGLE_CONTRIBUTION_BPS_THRESHOLD = 20000`; check is `amount * 10000 > funding_goal * 20000`). Also, this line conflates the single-contribution rule with "burst" — those are two separate triggers (the burst one is >10 contributions/ledger per campaign).

src/lib.rs
```diff
@@ -649,6 +656,8 @@ impl ProofOfHeart {
         decrement_active_campaign_count(&env);
         increment_cancelled_campaign_count(&env);
 
+        env.events()
+            .publish(("campaign_cancelled", campaign_id), campaign.amount_raised);
```
dmystical-coder: This adds a **second** `campaign_cancelled` emission — the original on the next line already carries `amount_raised` as data. Indexers will now see two cancellation events per campaign (different topic shapes). Please keep just one. If the goal was to add `amount_raised`, it was already present in the existing event's data.

src/lib.rs
```diff
@@ -1044,10 +1056,14 @@ impl ProofOfHeart {
         assert_admin(&env, &admin)?;
         bump_instance_ttl(&env);
         env.storage().instance().set(&DataKey::Paused, &false);
+        env.storage().instance().set(&DataKey::AutoPaused, &false);
```
dmystical-coder, line 1057: This is the core #409 fix and it's reasonable. Caveat: because `contribute()` sets `AutoPaused` and then returns `Err` in the same call, the write is rolled back and `AutoPaused` never persists in production (see `tests/test_contribute_caps.rs:87-88`). So this clears a flag that the normal flow can't actually set — defensive, but the underlying auto-pause mechanism needs a separate fix to be meaningful. Worth raising with maintainers.

src/test.rs
```diff
@@ -4132,3 +4082,88 @@ fn test_verify_campaigns_partial_failure_returns_err() {
...
+    client.verify_campaign(&campaign_id);
+
+    let res = client.try_contribute(&campaign_id, &contributor1, &2001i128);
+    assert_eq!(res.unwrap_err().unwrap(), Error::ContractPaused);
+
+    let _ = token;
+}
+
+#[test]
+fn test_unpause_clears_auto_pause_when_resume_campaign_blocked() {
```
dmystical-coder, line 4112: This test doesn't prove what its name claims. After the "trigger" at line ~4133 the contribution returns `Err` and `AutoPaused` is already rolled back, so the subsequent `unpause()` is a no-op and `contribute(500)` would succeed regardless of this PR. Likewise `resume_campaign` returns `CampaignNotActive` from the `require_active_campaign` check independent of pause state. To genuinely exercise recovery from a stuck auto-pause, set `DataKey::AutoPaused = true` directly via `env.as_contract(...)` (as `issues_test.rs` does) and then assert a state-changing call is blocked *before* `unpause()` and works *after*.

dmystical-coder (changes requested):
Thanks for taking this on, @Skinny001 — the formatting pass and the `test_admin.rs` cleanup (dropping the contradictory unreachable assertion) are nice, and the build is green (345 tests pass, clippy clean). However, I'm requesting changes: as written the PR only partially satisfies #409 and #410, and it introduces a duplicate event. Details below, with inline notes on the exact lines.

## #409 — `unpause()` clearing `AutoPaused`
The one-line fix matches the issue's proposed solution and is a sensible defensive change. But please be aware of a deeper problem the fix sits on top of: **`AutoPaused` can never actually persist in production.** Both auto-pause branches in `contribute()` do `set(AutoPaused, true)` and then `return Err(ContractPaused)` in the same invocation — and Soroban rolls back all storage writes (and discards events) when a call returns `Err`. The repo's own tests already encode this: `src/tests/test_contribute_caps.rs:87-88` asserts `!client.is_paused()` right after a trigger, with the comment *"Rollback ensures it's NOT paused."*

Consequences:
- The #409 "contract becomes a brick" scenario is **not reproducible** through `contribute()`, because the flag never sticks.
- The acceptance test #409 asks for ("trigger auto-pause via a burst contribution … then verify operations work again") can't be written truthfully — operations were never blocked. The new `test_unpause_clears_auto_pause_when_resume_campaign_blocked` test passes for the wrong reasons (see inline note).

Please either (a) flag this to the maintainers so the auto-pause mechanism itself is fixed first (set the flag in a path that returns `Ok`), or (b) note in the PR/issue that this is a defensive-only change and adjust the test to set `AutoPaused` directly (as the existing `issues_test.rs` helpers do) so it actually exercises a stuck→recovered state.

## #410 — event payload docs (largest gap)
Two blocking issues:

1. **Wrong file.** Issue #410 is about the existing `EVENT_PAYLOADS.md` at the **repo root**. This PR creates a brand-new `docs/EVENT_PAYLOADS.md` and leaves the root file untouched, so the repo now has two competing event docs and the one the issue targets is still wrong. Please either update the root file in place, or delete it and make `docs/EVENT_PAYLOADS.md` canonical (and update any links). Confirm the intended location with the maintainers.

2. **Incomplete + inaccurate**, so the acceptance criteria aren't met:
   - #410 requires *every* `publish(...)` in `src/lib.rs`/`src/voting.rs` to have a section. The new doc covers 24 events; at least these 16 emitted events are still undocumented — and they overlap heavily with the list the issue explicitly called out:
     `admin_transfer_initiated`, `admin_transfer_cancelled`, `campaign_deadline_extended`, `campaign_fee_override_set`, `category_duration_cap_set`, `category_duration_cap_removed`, `creation_disabled_updated`, `migrated`, `reserve_withheld`, `revenue_pool_refunded`, `token_update_proposed`, `token_update_accepted`, `token_update_cancelled`, `vesting_params_updated`, `voting_state_purged`, `warning_high_voting_balance`.
   - #410 requires documented payloads to match the emit shape. At least `auto_paused` and `campaign_resumed` are documented with the wrong topics/data (inline notes).

## Bug introduced — duplicate `campaign_cancelled` event
`cancel_campaign()` now emits the event twice (inline note on `src/lib.rs:660`). Indexers will see two cancellations per campaign. Please keep a single emission.

## Minor
- The PR description says it "adds `resume_campaign()` and auto-pause trigger" — both already exist on `main`; this PR doesn't add them. Worth correcting so reviewers know what's actually new.
- New tests use awkward `let (env, ...) = { let (e, ...) = setup_env(); (...) };` re-destructuring and `let _ = token;`; just bind `_token` in the pattern.

Happy to re-review once the doc location/coverage and the duplicate event are sorted. Thanks again!

Skinny001:
> Thanks for taking this on, @Skinny001 — the formatting pass and the `test_admin.rs` cleanup (dropping the contradictory unreachable assertion) are nice, and the build is green (345 tests pass, clippy clean). However, I'm requesting changes: as written the PR only partially satisfies #409 and #410, and it introduces a duplicate event. Details below, with inline notes on the exact lines.
> 
> ## #409 — `unpause()` clearing `AutoPaused`
> The one-line fix matches the issue's proposed solution and is a sensible defensive change. But please be aware of a deeper problem the fix sits on top of: **`AutoPaused` can never actually persist in production.** Both auto-pause branches in `contribute()` do `set(AutoPaused, true)` and then `return Err(ContractPaused)` in the same invocation — and Soroban rolls back all storage writes (and discards events) when a call returns `Err`. The repo's own tests already encode this: `src/tests/test_contribute_caps.rs:87-88` asserts `!client.is_paused()` right after a trigger, with the comment _"Rollback ensures it's NOT paused."_
> 
> Consequences:
> 
> * The #409 "contract becomes a brick" scenario is **not reproducible** through `contribute()`, because the flag never sticks.
> * The acceptance test #409 asks for ("trigger auto-pause via a burst contribution … then verify operations work again") can't be written truthfully — operations were never blocked. The new `test_unpause_clears_auto_pause_when_resume_campaign_blocked` test passes for the wrong reasons (see inline note).
> 
> Please either (a) flag this to the maintainers so the auto-pause mechanism itself is fixed first (set the flag in a path that returns `Ok`), or (b) note in the PR/issue that this is a defensive-only change and adjust the test to set `AutoPaused` directly (as the existing `issues_test.rs` helpers do) so it actually exercises a stuck→recovered state.
> 
> ## #410 — event payload docs (largest gap)
> Two blocking issues:
> 
> 1. **Wrong file.** Issue #410 is about the existing `EVENT_PAYLOADS.md` at the **repo root**. This PR creates a brand-new `docs/EVENT_PAYLOADS.md` and leaves the root file untouched, so the repo now has two competing event docs and the one the issue targets is still wrong. Please either update the root file in place, or delete it and make `docs/EVENT_PAYLOADS.md` canonical (and update any links). Confirm the intended location with the maintainers.
> 2. **Incomplete + inaccurate**, so the acceptance criteria aren't met:
>    
>    * #410 requires _every_ `publish(...)` in `src/lib.rs`/`src/voting.rs` to have a section. The new doc covers 24 events; at least these 16 emitted events are still undocumented — and they overlap heavily with the list the issue explicitly called out:
>      `admin_transfer_initiated`, `admin_transfer_cancelled`, `campaign_deadline_extended`, `campaign_fee_override_set`, `category_duration_cap_set`, `category_duration_cap_removed`, `creation_disabled_updated`, `migrated`, `reserve_withheld`, `revenue_pool_refunded`, `token_update_proposed`, `token_update_accepted`, `token_update_cancelled`, `vesting_params_updated`, `voting_state_purged`, `warning_high_voting_balance`.
>    * #410 requires documented payloads to match the emit shape. At least `auto_paused` and `campaign_resumed` are documented with the wrong topics/data (inline notes).
> 
> ## Bug introduced — duplicate `campaign_cancelled` event
> `cancel_campaign()` now emits the event twice (inline note on `src/lib.rs:660`). Indexers will see two cancellations per campaign. Please keep a single emission.
> 
> ## Minor
> * The PR description says it "adds `resume_campaign()` and auto-pause trigger" — both already exist on `main`; this PR doesn't add them. Worth correcting so reviewers know what's actually new.
> * New tests use awkward `let (env, ...) = { let (e, ...) = setup_env(); (...) };` re-destructuring and `let _ = token;`; just bind `_token` in the pattern.
> 
> Happy to re-review once the doc location/coverage and the duplicate event are sorted. Thanks again!

Done

dmystical-coder (approved):
Re-reviewed at `21ba527`. All seven points from the previous review are addressed — thanks for the thorough turnaround, @Skinny001. Verified locally: build is green, **345 tests pass, clippy clean**, and every `publish(...)` event in the code now has a matching doc section (0 missing).

**Resolved:**
- ✅ **#410 wrong file** — `docs/EVENT_PAYLOADS.md` deleted; the root `EVENT_PAYLOADS.md` (the file the issue names) is rewritten in place.
- ✅ **#410 coverage** — all 48 publish sites documented (the 16 previously-missing events are now present). I spot-checked `campaign_vote_cast` `(approve, balance, weight)`, `admin_updated`, `admin_transfer_initiated`, `campaigns_bulk_verified`, and `reserve_withheld` against the code — payloads match.
- ✅ **#410 wrong payloads** — `auto_paused` now correctly shows `("auto_paused",)` topic with `("huge_contribution", amount)` / `("burst", count)` data (split into two entries); `campaign_resumed` now includes the `caller` topic.
- ✅ **Duplicate `campaign_cancelled`** — second emission removed (`lib.rs`), and the `test_cancel_and_refund` snapshot updated to drop the extra event.
- ✅ **CAMPAIGN_LIFECYCLE threshold** — now states the 200% rule with the formula, separates the burst trigger, and documents the rollback caveat.
- ✅ **#409 test** — `test_unpause_clears_auto_pause_when_resume_campaign_blocked` now sets `AutoPaused` directly, asserts a call is blocked *before* `unpause()` and works *after* — it genuinely exercises recovery now.
- ✅ Test tuple destructuring cleaned up; `let _ = token` removed.

**Two non-blocking notes for follow-up (your call / maintainers'):**
1. The doc footer says "48 documented publish() call sites" but there are 47 `###` sections (a couple of events emit from two sites, e.g. `admin_transfer_cancelled`). Minor wording nit.
2. The underlying point stands that `AutoPaused` can't persist via `contribute()` (the `Err` rolls back the write) — so the production auto-pause safety net is effectively inert. That's pre-existing and out of scope here; the docs now honestly call it out. Worth a separate issue if the team wants auto-pause to actually engage.

LGTM to merge once #409/#410 owners are happy. Nice work.

PR #568, opened 2026-06-30
Use effective raised amount for withdrawals

### LINKED ISSUE #514: [Bug] withdraw_funds fee calculation uses amount_raised not effective_amount_raised — overpays platform fee after refunds
## Problem
If some contributors claimed refunds before the creator withdrew (e.g. during the funded-not-withdrawn window), `amount_raised` is higher than `effective_amount_raised`. Using `amount_raised` for fee calculation means the platform takes fees on money that was already refunded.

## Fix
Use `campaign.effective_amount_raised` as the basis for fee and creator payout calculations in `withdraw_funds`.

Killerjunior:
## 📌 Description
Fixes `withdraw_funds` so platform fees, creator payout, reserve calculation, and global raised cleanup use `campaign.effective_amount_raised` instead of historical `campaign.amount_raised`. This prevents fees from being charged on funds that have already been refunded.

## 🔗 Related Issues
Fixes #514

## 🧪 Changes Made
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation update

## ✅ Checklist
- [ ] Code compiles successfully
- [x] Documentation updated (if required)

## ⚠️ Breaking Changes
None.

Not applicable.

## 🧩 Additional Notes
Added a regression test proving withdrawal fees are calculated from live effective funds after a simulated refund. I could not run the focused tests locally because Cargo dependency resolution failed on crates.io SSL/cached dependency availability for `ryu`; `cargo fmt --all` completed successfully.


davidmaronio (changes requested):
the intent (use the post-refund effective amount as the fee/reserve basis) is reasonable, but it doesn't compile, clippy: E0425 cannot find value `withdraw_basis` (withdraw.rs, x2). you reference withdraw_basis but never bind it, add `let withdraw_basis = campaign.effective_amount_raised;`. then run cargo fmt and rebase onto a clean main (the other compile errors in the log are from a stale base, not your diff). note #583 carries this same withdraw change plus a coverage job, so one of the two should drop it.

PR #585, opened 2026-07-10
feat: Add get_campaign_stats query

Emelie-Dev:
closes #538


davidmaronio (changes requested):
thanks for taking another swing at the stats query, and credit where due: unlike the earlier attempt at this issue (#575) you did not touch the existing DataKey encoding, which was the dealbreaker there. the CampaignStats struct shape is sensible too.

that said, this doesn't work as submitted, and the CI red is from this branch, not the base:

1. src/storage.rs:396 and :415: get_top_contributor / set_top_contributor / get_last_contribution_time / set_last_contribution_time reference DataKey::TopContributor and DataKey::LastContributionTime, but those variants don't exist in the DataKey enum on main and this PR doesn't add them. this can't compile. if you add them, append them at the END of the enum so existing key encodings are untouched (this is exactly what sank #575).
2. missing the actual feature: the title says "add get_campaign_stats query" but there's no get_campaign_stats entry point in src/lib.rs, nothing computes avg_contribution, and nothing ever calls set_top_contributor / set_last_contribution_time from the contribute path. as it stands the stats would always be empty. wire the setters into contribute() and expose the query.
3. src/types.rs:205: top_contributor is typed MaybePendingCreator but your storage helper returns Option<Address>. pick one, Option<Address> seems more natural for a contributor.
4. out of scope: the .github/workflows/ci.yml toolchain pinning to 1.82.0, the Cargo.toml dev-dependency pin, and the deleted trailing newline on Cargo.toml belong in a separate PR if they're needed at all. please revert them here.
5. the PR is also behind main, so rebase while you're at it.

gate to re-review: a rebased branch where the contract compiles, get_campaign_stats exists and is tested end to end (contribute then query), DataKey variants are appended not inserted, and the CI/Cargo changes are reverted.

PR #605, opened 2026-07-29
fix(campaigns): prevent bait-and-switch edits on verified campaigns

AbuJulaybeeb:
This PR resolves #456 by ensuring that campaign content cannot be edited once verified, regardless of whether it has received contributions.

Changes Included:

src/campaigns/update.rs: Added require_unverified_campaign(&campaign)? to update_campaign_description. (Note: The original issue description transposed the function names; update_campaign already had this guard, whereas update_campaign_description was missing it).
Acceptance Criteria Verified:

 Verified campaigns are strictly blocked from content updates.
 Applies consistently across both update_campaign and update_campaign_description.
 Linked issue correctly.
How to Test:

bash
cargo test

closes #456


davidmaronio (changes requested):
thanks for taking a shot at #456. good instinct moving `require_not_paused` before the campaign load, that matches the ordering in `update_campaign`. but a few blockers:

1. src/campaigns/update.rs:63 - the pr description says this adds `require_unverified_campaign(&campaign)?`, but the diff never calls it. what actually landed is a `funds_withdrawn` check, so a verified campaign that hasn't withdrawn can still edit its description. the core bait-and-switch issue is still open. please add the `require_unverified_campaign` guard as described.
2. src/admin.rs:536, src/lib.rs:340, src/storage.rs:82 - `set_category_max_goal_cap` / `get_category_max_goal_cap` and the new `CategoryMaxGoalCap` storage key are unrelated to this issue, and nothing enforces the cap in `create_campaign`, so it's write-only storage plus new public contract surface. this exact feature was part of an accidental merge that got reverted in #702. please drop it from this pr.
3. no tests cover the new behavior. a test that verifies a campaign and then asserts `update_campaign_description` fails would prove the fix.

heads up that #698 and #702 already implement the verified-edit guard on the same lines, so this will conflict with whichever merges first.

PR #609, opened 2026-07-29
fix(admin): reorder resume_campaign to check AutoPaused before require_active_campaign (#436)

OTimileyin:
## 📌 Description

Reorders the `resume_campaign` function to check the global `AutoPaused` flag **before** `require_active_campaign`. Previously, if the campaign that triggered the auto-pause was cancelled or withdrawn, `resume_campaign` could never clear the global flag because `require_active_campaign` would fail first. The admin recovery path via `unpause()` already clears both `Paused` and `AutoPaused` flags.

## 🔗 Related Issues

Closes #436

## 🧪 Changes Made
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation update

## ✅ Checklist
- [x] Tests added/updated and passing (all 8 resume/unpause tests pass)
- [x] Documentation updated (CHANGELOG.md)
- [x] No breaking changes

## ⚠️ Breaking Changes

None. The `unpause()` function already cleared `AutoPaused` alongside `Paused` — this change only reorders checks in `resume_campaign` for a faster early-return when not paused.

## 🧩 Additional Notes

Two existing tests (`test_unpause_clears_auto_pause_when_resume_campaign_blocked` in both `test_campaigns.rs` and `test_campaign_update.rs`) were updated to expect `ValidationFailed` instead of `CampaignNotActive` when calling `resume_campaign` after `unpause()`. This is correct: `unpause()` clears `AutoPaused`, so the new early check correctly returns `ValidationFailed` (no pause active).


davidmaronio (approved):
nice clean fix. the reorder in src/admin.rs:507 is exactly what #436 called for, require_auth stays first so there's no auth change, and updating the two unpause tests to expect ValidationFailed with an explanation of why is the right call rather than papering over it.

two small things, neither blocking:

1. src/tests/test_campaign_update.rs:59 - the variable extraction and #349 comments on the event tests are unrelated to this fix (and the same hunks appear in your #610). fine either way since they're identical, but ideally keep each pr to its own change.
2. worth noting in the changelog entry that when the triggering campaign is cancelled, resume_campaign still can't clear the flag (require_active_campaign runs before the write), so unpause() remains the recovery path. the entry currently reads like both paths are fixed.

PR #610, opened 2026-07-29
fix(storage): remove dead BlockContributionCount key and unused get/set helpers (#435)

OTimileyin:
## 📌 Description

Removes the dead `BlockContributionCount` variant from the `ContributionKey` enum and deletes its unused `get_block_contribution_count` and `set_block_contribution_count` helpers. Only the per-campaign `BlockCampaignContributionCount` variant is actually read/written by the `contribute()` anomaly-detection burst guard.

## 🔗 Related Issues

Closes #435

## 🧪 Changes Made
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation update

## ✅ Checklist
- [x] Tests added/updated and passing (394 pass, 2 pre-existing failures on main unrelated)
- [x] Documentation updated (CHANGELOG.md)
- [x] No breaking changes (variant was never read or written by any caller)

## ⚠️ Breaking Changes

None. The `BlockContributionCount` variant was never written by any caller — `contribute()` only uses the per-campaign `BlockCampaignContributionCount`. No ledger entries exist for this key.

## 🧩 Additional Notes

The `ROADMAP.md` already tracked this as "🟡 P2 — BlockContributionCount (global) is dead storage." This PR fully resolves that item.


davidmaronio (approved):
straightforward and correct. the variant was never written by any caller and both helpers were already flagged `#[allow(dead_code)]`, so this is pure cleanup with no ledger-state risk, and the changelog entry under removed is appreciated.

one small ask:

1. src/tests/test_campaign_update.rs:59 - the event-test refactor and #349 comments aren't related to #435 (they also appear in your #609). would keep this pr to just storage.rs and the changelog.

also flagging that #690 is an open duplicate with the identical storage.rs change, so we'll pick one and close the other.

PR #611, opened 2026-07-29
fix: snapshot vesting params per-campaign at creation to prevent retroactive changes (#466)

sojetunde8:
Closes #466

## Summary
- Vesting parameters (reserve_bps, delay_days) are now snapshotted into per-campaign storage at creation time via a new `CampaignVesting(u32)` key
- `withdraw_funds` reads the per-campaign snapshot when available, falling back to global defaults for pre-upgrade campaigns (backward compatible)
- A regression test verifies that changing `set_vesting_params` after campaign creation does not affect already-created campaigns

## Testing
- All 15 withdrawal tests pass, including the new regression test `test_vesting_snapshot_not_affected_by_later_changes`
- All 47 admin tests pass
- All 14 lifecycle tests pass
- `cargo check` compiles cleanly with no errors or warnings

### DIFF
--- a/src/campaigns/create.rs
+++ b/src/campaigns/create.rs
@@ -6,10 +6,12 @@ use crate::storage::{
     bump_instance_ttl, get_campaign_count, get_category_campaign_bucket,
     get_category_campaign_count, get_category_duration_cap, get_creation_disabled,
     get_creator_campaign_bucket, get_creator_campaign_count, get_max_campaign_funding_goal,
-    get_min_campaign_funding_goal, set_campaign, set_campaign_count, set_campaign_creator_index,
-    set_campaign_start_time, set_category_campaign_bucket, set_category_campaign_count,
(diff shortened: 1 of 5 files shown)


davidmaronio (approved):
nice clean fix for #466. snapshotting both delay and reserve bps at creation with a fallback for pre-upgrade campaigns is exactly the right shape, and the regression test checking exact payout math on both a pre-vesting and post-vesting campaign is solid.

one small thing:

1. src/storage.rs:902 - `remove_campaign_vesting` is dead code behind `#[expect(dead_code)]`. campaigns are never deleted so nothing will ever call it. i'd just drop it, we can add it back if a purge path ever needs it.

also noting for other reviewers: the benchmark test drop from 70 to 58 campaigns in test_benchmark.rs is explained in the comment there (testutils host limit from the extra per-campaign entry) and still covers the bucketed pagination page, so i'm fine with it.

PR #612, opened 2026-07-29
docs: document admin trust model for single-key control (#468)

sojetunde8:
Closes #468

## Summary
- Added comprehensive Admin Power Concentration section to docs/THREAT_MODEL.md documenting all sensitive operations controlled by the single admin key
- Documented realistic attack vectors for a compromised admin key
- Provided risk management recommendations including key hygiene, monitoring, transparency commitments, and future improvement paths
- Added reference to the admin trust model in SECURITY.md with key implications for users
- Added Token Migration Timelock subsection documenting the existing 7-day delay safeguard

## Testing
- Documentation-only change - no code modified
- Cross-referenced links between SECURITY.md and THREAT_MODEL.md are correct

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -80,8 +80,19 @@ Community verification uses a token-gated voting model:
 Security assumptions and limitations:
 
 - This mechanism is **not inherently sybil-resistant**: a single token holder can split tokens across many addresses to inflate the _vote count_ and reach quorum more easily (even though total voting weight stays similar).
-- The model assumes the token's distribution and issuance are outside the contract’s control; if token minting is centralized or cheaply obtainable, governance can be captured.
+- The model assumes the token's distribution and issuance are outside the contract's control; if token minting is centralized or cheaply obtainable, governance can be captured.
 - Admin verification (`verify_campaign`) is a privileged path; users should treat the stored admin as a trust assumption for campaign verification.
+
+## Admin Trust Model (#468)
+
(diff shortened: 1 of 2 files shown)


davidmaronio (changes requested):
this is a genuinely useful writeup and clearly written from the actual code, the entrypoint list, event names, and the accept_token_update zero-balance guard all check out. a few accuracy fixes before merge since this is a security doc:

1. docs/THREAT_MODEL.md (fee override bullet + attack vector 2) - says fees can go up to PLATFORM_FEE_ABSOLUTE_MAX_BPS (100%). both update_platform_fee and set_campaign_fee_override also enforce PLATFORM_FEE_MAX_BPS = 1000, so the real ceiling is 10%. please correct the range, and keep the 0% part which is accurate.
2. docs/THREAT_MODEL.md (token migration timelock section) - the delay is not fixed at 7 days anymore. propose_token_update uses get_token_update_delay_secs with TOKEN_UPDATE_DELAY_SECS as the default, and the admin can change it via set_token_update_delay_secs (#650). worth stating explicitly that the timelock itself is admin-configurable, which weakens it as a defense against a compromised admin key, that is exactly the kind of nuance this doc exists for.
3. docs/THREAT_MODEL.md (monitoring bullet) - campaign_created is emitted by any creator, not by admin actions. drop it from the admin-events list.
4. optional: the admin transfer bullet only mentions the two-step initiate/accept flow, but update_admin(new_admin) also exists as a single-step path and probably deserves a mention.

on the overlap with #691: this pr fits the existing THREAT_MODEL/SECURITY structure and is verified against real entrypoints, so i'd take this one and close #691 as duplicate once the fixes above land.

PR #615, opened 2026-07-29
fix: cancel_campaign must decrement total_raised_global upfront (#439)

Bidex-cmyk:
closes #439 
Decrement total_raised_global immediately in cancel_campaign and admin_cancel_campaign so cancelled campaigns with unclaimed refunds no longer inflate global stats.

Also introduces PendingRefundTotal storage counter to track unclaimed refunds across cancelled campaigns, so accept_token_update can reliably block when escrowed refunds remain (even after total_raised_global is zeroed at cancel time).

Changes:
- Add PendingRefundTotal to AdminKey storage (track unclaimed refunds on cancelled campaigns)
- Subtract campaign.amount_raised from total_raised_global upfront in cancel_campaign and admin_cancel_campaign
- Increment pending_refund_total at cancel time
- In claim_refund, decrement pending_refund_total for cancelled campaigns (skip total_raised_global to avoid double-decrement)
- In accept_token_update, gate on pending_refund_total == 0 in addition to total_raised_global == 0
- Initialize pending_refund_total to 0 in init()

Co-authored-by: Promise Olubudo <254755326+baedboibidex-cmyk@users.noreply.github.com>


davidmaronio (changes requested):
the core design here is sound and thoughtful: decrementing `total_raised_global` upfront in `cancel_campaign` / `admin_cancel_campaign` (src/campaigns/cancel.rs), moving the escrow into a new `PendingRefundTotal` counter, decrementing that per-claimant in `claim_refund` (src/contributions.rs:285), and widening the `accept_token_update` gate in src/admin.rs:365 to include the pending refund total keeps the token-swap guard honest without double-counting. updating the `get_platform_stats` expectation with a comment explaining why is also the right move.

things to fix:

1. scope: the diff carries a lot that has nothing to do with #439 - the whole `remove_personal_cap` entrypoint (#503) with its tests, error variant, CHANGELOG/EVENT_PAYLOADS/AUTHORIZATION doc entries, the duplicate-contract revert (src/proof_of_heart deletions, voting_tests.rs, frontend file deletions), and the `accept_campaign_transfer` pause reorder. some of this may be base drift since the branch is conflicted, but as it stands the PR is unreviewable as one unit. please rebase onto current main so the diff is only the #439 change.
2. migration hazard worth addressing (or at least documenting): a campaign cancelled before this change deploys has its escrow still counted in `total_raised_global` and nothing in `PendingRefundTotal`. after upgrade, `claim_refund` for that campaign takes the `is_cancelled` branch and does `pending_refund.checked_sub(amount)` on a zero counter, which returns `Err(Overflow)`... actually checked_sub on i128 only errors on true overflow, so it would drive `PendingRefundTotal` negative and leave `total_raised_global` permanently inflated, blocking token swaps forever. a `migrate` step or a floor-at-zero with an explicit comment is needed.
3. src/campaigns/cancel.rs - the same two-counter block is pasted into both `cancel_campaign` and `admin_cancel_campaign`. pull it into a small helper so the two paths can't drift.
4. overlap: #714, #604 and #718 touch the same accounting. worth agreeing with the maintainers which approach wins before polishing further.

gate: this branch has merge conflicts with main. please resolve them (which should also shed the unrelated changes), address point 2, and push.

PR #619, opened 2026-07-30
fix: validate set_personal_cap against lifetime_contribution (#441)

CHKM001:
PR #441 Validate `set_personal_cap` against `lifetime_contribution`

## Summary

`set_personal_cap_fn` did not validate the new cap against the caller's existing
`lifetime_contribution`. A contributor could set their cap to 1 stroop after
having contributed 1000 XLM, permanently blocking further contributions with no
way to recover.

## Change

Added a check at `src/contributions.rs:312-315` that rejects any cap lower than
the contributor's current `lifetime_contribution`:

```rust
let lifetime = get_lifetime_contribution(env, campaign_id, &contributor);
if amount < lifetime {
}
```

This ensures the personal cap always stays at or above what the contributor has
already put in, preventing accidental or malicious self-freeze.

Closes #441


davidmaronio (changes requested):
the core fix is right: rejecting a personal cap below the caller's lifetime_contribution in src/contributions.rs closes the self-freeze described in #441, and you added a regression test in src/tests/test_regressions.rs. good instinct.

but the branch as pushed can't be merged:

1. the diff deletes large chunks of the repo that have nothing to do with this fix: src/admin.rs (-35), src/clients.rs (-28), src/events.ts (-32), src/types.ts (-20), src/proof_of_heart/src/admin.rs (-36), src/proof_of_heart/src/voting.rs (-33), src/tests/voting_tests.rs (-22), src/test.rs (-23), src/campaigns.rs (-7), plus -31 lines from src/lib.rs. this looks like the branch was cut from a stale or foreign snapshot of the repo, so merging it would wipe current code. please recreate the branch from current main and cherry-pick only the contributions.rs check and the new test.
2. the PR is DIRTY and CI is red, both of which should resolve themselves once the branch is rebuilt on main with just the 4-line fix and the test.
3. one behavior question to settle in the reworked version: should setting a cap exactly equal to lifetime_contribution be allowed (it is with `amount < lifetime`)? that seems fine since it just blocks future contributions without stranding anything, but add a test asserting the equal case so the boundary is intentional.

also note #688 is touching the neighboring cap validation from #457, so keep the diff tightly scoped to avoid colliding with it. gate for merge: a fresh branch off main containing only the contributions.rs validation and its tests, with green CI. without that within a week or two we'll close this as stale.

PR #682, opened 2026-07-30
fix: correct revenue_deposited event topics in docs and add topic ver…

shogun444:
## Summary

This PR fixes the `revenue_deposited` event documentation to match the contract's actual behavior and adds regression coverage to ensure the documented event structure remains accurate.

The contract already publishes the `revenue_deposited` event with a three-field topic tuple containing the event name, campaign ID, and creator address. However, `EVENT_PAYLOADS.md` documented only a two-field tuple, which could cause indexers or downstream consumers to parse the event incorrectly.

### Updated Event Documentation

- Corrected the `revenue_deposited` event topics in `EVENT_PAYLOADS.md` to include the creator address.
- Updated the source reference to the current implementation in `src/revenue.rs`.

### Added Regression Test

- Added `test_deposit_revenue_event_includes_creator`.
- Verified that the published event contains:
  - `"revenue_deposited"`
  - The correct `campaign_id`
  - The correct creator `Address`
- Added regression coverage to ensure the documented event topic structure remains in sync with the contract implementation.

## Notes

- No contract behavior was changed.
- The implementation already emitted the correct event topics. This PR updates the documentation to reflect the existing implementation and adds a regression test to prevent future documentation drift.

## Verification

- `cargo fmt --check`
- `cargo clippy --all-targets --features testutils -- -D warnings`
- `cargo test --features testutils`

### Results

- `cargo fmt --check` passed.
- `cargo clippy --all-targets --features testutils -- -D warnings` passed.
- `395/397` tests passed (`2` pre-existing unrelated test failures).

Closes #673

### DIFF
--- a/EVENT_PAYLOADS.md
+++ b/EVENT_PAYLOADS.md
@@ -158,9 +158,9 @@ Every `publish(...)` call in the contract, with its topics, data shape, and the 
 
 | Field   | Value                                                      |
 |---------|------------------------------------------------------------|
-| Topics  | `("revenue_deposited", campaign_id: u32)`                  |
+| Topics  | `("revenue_deposited", campaign_id: u32, creator: Address)`|
 | Data    | `amount: i128`                                             |
-| Source  | `lib.rs:842` — `deposit_revenue()`                         |
+| Source  | `src/revenue.rs:44` — `deposit_revenue()`                  |
 
 ---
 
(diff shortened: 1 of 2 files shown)


davidmaronio (approved):
rebase looks clean, doc and test match the publish call on main. merging.

PR #683, opened 2026-07-30
fix(revenue): add upper bound check in deposit_revenue

Peolite1:
## 📌 Description
This PR addresses the issue where `deposit_revenue` had no ceiling on the deposited amount, which could lead to an oversized pool that triggers a `checked_mul` overflow error much later during `claim_revenue`. 
It introduces an upfront upper-bound check in `deposit_revenue` to fail gracefully before any state changes are made if the pool gets too large. It also adds the `test_deposit_revenue_too_large` unit test.

## 🔗 Related Issues
Closes #674

## 🧪 Changes Made
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation update

## ✅ Checklist

## ⚠️ Breaking Changes
None.

N/A

## 🧩 Additional Notes
N/A


davidmaronio (changes requested):
the core fix is good: checking `effective_amount_raised * new_pool * revenue_share_percentage` up front in `deposit_revenue` fails the worst case before any state change or transfer, and `test_deposit_revenue_too_large` (with the i128::MAX / 1_000_000 mint to keep the mock token happy) exercises it properly.

the PR needs slimming down before it can merge though:

1. pr_body.txt: this is a stray scratch file (it even contains the body of your other PR, #686). please delete it from the branch.
2. src/queries.rs:126 and :176: dropping the `bucket_len == 0` branch in `get_campaigns_by_category` / `get_creator_campaigns` is an unrelated behavior change to pagination that isn't mentioned anywhere in the PR. please leave queries.rs untouched here; if you think that logic is wrong, open a separate issue/PR for it.
3. src/admin.rs, src/campaigns.rs, src/lib.rs, src/campaigns/withdraw.rs, src/tests/test_campaign_update.rs: these hunks are cleanup of the broken #618 leftovers, which main has since fixed itself, which is why the PR is now showing conflicts. after a rebase these should all disappear.

so the gate is: drop pr_body.txt and the queries.rs change, rebase onto current main (`git fetch upstream && git rebase upstream/main && git push --force-with-lease`) so the diff is just revenue.rs + test_revenue_deposit.rs, and let CI go green.

one more thing worth flagging: this branch, #685, and #686 are stacked on each other (each contains the previous one's changes) and come from two accounts, Peolite001 and Peolite1. if those are both you, please pick one account and keep each PR scoped to a single issue, it makes review much easier.

PR #685, opened 2026-07-30
Issue 678 revenue test

Peolite1:
## 📌 Description
This PR updates the `deposit_revenue` event test to strictly assert the exact tuple payload emitted by the contract. This prevents any future drift where the event topic tuple shape might change without the test failing, ensuring it always stays perfectly aligned with `EVENT_PAYLOADS.md`.

## 🔗 Related Issues
Resolves #678

## 🧪 Changes Made
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [x] Documentation update (Tests updated to enforce exact event payload specification)

## ✅ Checklist

## ⚠️ Breaking Changes
None.

N/A

## 🧩 Additional Notes
The test now constructs the exact expected tuple `("revenue_deposited", campaign_id, creator)` and compares it directly against the emitted event's topic tuple, rather than just iterating and extracting fields individually.


davidmaronio (changes requested):
the actual #678 work here is good: `test_revenue_deposited_event_payload` builds the exact expected topic tuple `("revenue_deposited", campaign_id, creator)` and compares it whole, which is exactly the drift protection the issue asked for, and the EVENT_PAYLOADS.md correction to include the creator topic matches what the contract emits.

but most of the diff isn't this PR:

1. src/revenue.rs, src/tests/test_revenue_deposit.rs (`test_deposit_revenue_too_large`): this is #683's deposit upper-bound change, carried in wholesale. this branch is stacked on #683, so as written the two PRs can't be reviewed or merged independently. please rebase this branch onto main so it contains only the event test + EVENT_PAYLOADS.md.
2. src/queries.rs:126 and :176: same unrelated pagination change as in #683, please drop it here too.
3. src/admin.rs, src/campaigns.rs, src/lib.rs, src/tests/test_campaign_update.rs: leftovers of the #618 cleanup that main has since done itself, which is why the PR shows conflicts. they should vanish on rebase.

gate: rebase onto current main so the diff is just EVENT_PAYLOADS.md + the new event payload test (`git fetch upstream && git rebase upstream/main && git push --force-with-lease`), then CI green. happy to approve quickly after that, the test itself is ready.

PR #689, opened 2026-07-30
fix: use 1-address-1-vote model to prevent flash-loan voting (#469)

0dillon:
Closes #469

## Summary
Replace token-weighted voting with a **1-address-1-vote** model to close the flash-loan attack vector where an attacker borrows a large token balance, votes with inflated weight, and returns the tokens before verification.

## What Changed
- **`cast_vote`**: Each vote now contributes exactly `1` to the weight sum instead of the voter's live token balance. The token-holder gate (`min_voting_balance`) is preserved — you still need tokens to vote, but your balance doesn't affect your voting power.
- **`verify_with_votes`**: Threshold calculation now uses vote **counts** (`approve_votes` / `reject_votes`) instead of token **weights**. This is safe because the quorum check already guarantees `total_votes > 0`.
- **Backward compat**: `approve_weight` and `reject_weight` storage keys are still incremented (by 1 per vote) so existing campaigns with stored voting state remain valid.

## Design Decision
Three options were considered from the issue: snapshot at creation, minimum holding period, or 1-address-1-vote. **1-address-1-vote** was chosen because it's the simplest, most gas-efficient fix that fully eliminates the flash-loan vector without requiring expensive on-chain iteration of all token holders at campaign creation time.

## Acceptance Criteria
- [x] `cast_vote` no longer uses live balance for voting weight
- [x] `verify_with_votes` threshold immune to flash-loan inflation
- [x] Token-holder gate preserved (must hold >= `min_voting_balance`)
- [x] All tests pass (393/395, 2 pre-existing unrelated failures in `test_campaign_update`)
- [x] Clippy clean, formatting clean
- [x] Proptests updated for 1-address-1-vote invariant

## Test Output
```
test result: ok. 393 passed; 2 failed (pre-existing in test_campaign_update)
```

## Follow-ups (out of scope)
- Consider removing `approve_weight`/`reject_weight` storage entirely since they now equal vote counts (backward-compat cleanup)
- Update `EVENT_PAYLOADS.md` to reflect changed event shape

## Security Note
This fix prevents flash-loan governance attacks. The 1-address-1-vote model is Sybil-resistant only to the extent that obtaining token balances across many addresses costs real capital. The `min_voting_balance` admin parameter can be raised to increase this cost.

### DIFF
--- a/src/tests/test_regressions.rs
+++ b/src/tests/test_regressions.rs
@@ -273,9 +273,13 @@ fn test_set_personal_cap_cannot_exceed_max_contribution_per_user() {
     assert_eq!(res2.unwrap_err().unwrap(), Error::ValidationFailed);
 }
 
-// ── #354 vote weight checked addition ──
+// ── #354 / #469 vote weight safe addition ──
+/// With the 1-address-1-vote model (#469), vote weight is incremented by 1
+/// regardless of token balance. Even when the stored weight is near i128::MAX,
+/// a vote still succeeds because we only add 1 (not the voter's balance),
+/// preventing flash-loan attacks that inflate voting weight.
 #[test]
-fn test_vote_weight_overflow_fails() {
+fn test_vote_weight_does_not_overflow_with_1_address_1_vote() {
     let (env, _admin, creator, contributor, _, _token, token_admin, client) = setup_env();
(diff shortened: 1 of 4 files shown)


davidmaronio (approved):
overflow test adapted to the count-based model and ci is green. merging, thanks for the fast fix.

PR #693, opened 2026-07-30
fix: add funds_withdrawn guard to withdraw_reserve (#459)

samjay8:
Closes #459

## Summary

- Adds defense-in-depth `funds_withdrawn == true` guard to `withdraw_reserve`, preventing a migration-planted reserve on a non-withdrawn campaign from being drained

## Testing

- All 13 existing withdrawal tests pass, including the full vesting/reserve flow
- `cargo fmt --check` passes on changed files


davidmaronio (changes requested):
good catch and a nicely minimal fix. gating `withdraw_reserve` on `campaign.funds_withdrawn` in src/campaigns/withdraw.rs:131 closes the migration-planted-reserve drain path, and putting the check before `require_auth` keeps the error deterministic. the comment explaining the threat model is appreciated.

two asks before this can land:

1. src/campaigns/withdraw.rs - there's no test. for a fund-safety guard this repo's bar is a regression test: plant a `CampaignReserve` on a campaign with `funds_withdrawn == false` (via `env.as_contract` + the storage setter, same pattern the other storage-poking tests use), call `try_withdraw_reserve`, and assert `ValidationFailed`; then confirm the normal withdraw-then-reserve flow still succeeds. without it, nothing stops this guard from being refactored away later.
2. minor: `Error::ValidationFailed` works, but `Error::NoFundsToWithdraw` (or `FundingGoalNotReached`) would tell a caller more about why. not a blocker, your call - just say why you picked ValidationFailed if you keep it.

also flagging: #727 adds a withdraw_reserve guard for #443 in the same function, so whichever merges second will need a rebase - worth coordinating.

gate: the branch is behind main (CI is green on the current head). please add the test, rebase onto main, and push.

davidmaronio (approved):
the regression test is exactly the shape i asked for, planted reserve rejected then the legit withdraw-reserve flow verified in the same test, and the comment explaining the ValidationFailed choice settles my error question. ci green, merging. nice work.

PR #570, opened 2026-06-30
security: replace pull_request_target with two-workflow pattern (fixes #472)

CodingAngel1:
## Summary

Resolves **#472** — the `pull_request_target` trigger in `.github/workflows/auto-review.yml` exposes `secrets.OWNER_PAT` and `secrets.GROQ_API_KEY` to code running in the context of fork PRs, enabling supply-chain attacks.

This PR removes the vulnerable workflow and replaces it with the [GitHub-recommended two-workflow pattern](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/).

---

### Root cause

`pull_request_target` is a privileged event: it always runs in the context of the **base repository**, giving the workflow full access to repository secrets, even when the triggering commit comes from a fork. GitHub's documentation explicitly warns:

> "If the `pull_request_target` event is used with the `pull_request` trigger, the workflow will have write access to the target repository's resources, including secrets."

### What the old workflow did

```yaml
on:
  pull_request_target:            # <- privileged event
    types: [opened, synchronize, ...]

permissions:
  contents: write
  pull-requests: write

env:
  GROQ_API_KEY: ${{ secrets.GROQ_API_KEY }}   # <- secret injected
  GH_TOKEN:     ${{ secrets.OWNER_PAT }}       # <- PAT injected
```

A malicious contributor could craft a PR that:
1. Modifies the workflow itself (before it was disabled) or any script invoked by it.
2. Causes the runner to exfiltrate `GROQ_API_KEY` and `OWNER_PAT` to an attacker-controlled endpoint — e.g., via a `curl` in a Makefile, a Cargo build script, or a modified shell helper.
3. Uses the leaked `OWNER_PAT` (which has admin-level PR write access) to approve, merge, or close arbitrary PRs, inject malicious commits, or pivot to other repositories the token can access.

The workflow was disabled (`.disabled` suffix) but the file remained in the repository; re-enabling it (intentionally or accidentally) would immediately reintroduce the vulnerability.

---

## The Fix — Two-Workflow Pattern

The core principle: **never mix untrusted code execution with secret access**.

### Workflow 1 — `pr-collector.yml` (unprivileged)

| Property | Value |
|---|---|
| Trigger | `pull_request` |
| Runs in context of | fork commit (untrusted) |
| Secrets available | **None** — only ephemeral `GITHUB_TOKEN` (read-only) |
[...]

PR #696, opened 2026-07-30
fix(ci): stop the AI review workflow from bypassing branch protection (#467)

Fury03:
Closes #467

## Summary

The privileged review workflow merged with `gh pr merge --admin` using
`OWNER_PAT`. `--admin` overrides branch protection outright, so any PR the Groq
model scored at `confidence >= 0.7` was merged whether or not CI had passed —
which is how the build with duplicate error discriminants reached `main`.

### `--admin` removed

`.github/workflows/pr-review.yml` now calls `gh pr merge --squash` with no
override. Branch protection is the final authority: if it requires approvals,
an up-to-date branch, or checks this job did not inspect, GitHub rejects the
merge and the step fails loudly. That is the intended behaviour.

### CI must be green before a merge is attempted

New `Verify CI is green` step, run before the merge and required by its `if:`.
It reads `statusCheckRollup` and `mergeable` for the PR and refuses the merge
unless:

- at least one check is reported on the head commit,
- every check has **completed**,
- none concluded `FAILURE` / `TIMED_OUT` / `CANCELLED` / `ACTION_REQUIRED` /
  `STARTUP_FAILURE` (and no legacy commit status is `FAILURE` / `ERROR`),
- `mergeable == MERGEABLE`.

Details worth calling out:

- **Fails closed.** Pending, missing, and failing checks are all "no merge".
  Nothing about this step can turn a red build into a merge.
- **No self-deadlock.** This workflow's own check run appears in the PR's
  rollup, so entries matching `github.workflow` are filtered out. Without that
  the gate would wait on itself until it timed out.
- **Handles both rollup shapes.** `statusCheckRollup` mixes `CheckRun` objects
  (`status` / `conclusion`) with `StatusContext` objects (`state`); both are
  evaluated. A null rollup is treated as "no checks" rather than erroring.
- **Bounded wait.** Checks are normally finished by the time a reviewer types
  the trigger phrase, but the step polls for up to 5 minutes (10 × 30s) for
  stragglers before giving up and skipping the merge.

A skipped merge now leaves a comment naming the reason instead of failing
silently or going red on a legitimately-blocked PR.

### Actions pinned to commit SHAs

Every `uses:` across all four workflows is pinned to a full SHA with the tag
retained as a trailing comment. This matters most in `pr-review.yml`, where a
retagged or compromised upstream action would run in a job holding `OWNER_PAT`
and `GROQ_API_KEY`.

| Action | Pin |
| ------ | --- |
| `actions/checkout` | `11d5960a326750d5838078e36cf38b85af677262` (v4.4.0) |
| `actions/upload-artifact` | `ea165f8d65b6e75b540449e92b4886f43607fa02` (v4.6.2) |
| `dawidd6/action-download-artifact` | `bf251b5aa9c2f7eeb574a96ee720e24f801b7c11` (v6) |
| `dtolnay/rust-toolchain` | `4cda84d5c5c54efe2404f9d843567869ab1699d4` (stable), `4fd1da8b0805d2d2e936788875a7d65dbd677dc2` (nightly) |

Each pin is the current head of the major version already in use, so no
behaviour changes — only the reference is made immutable. `ci.yml` and
`fuzz.yml` are included because pinning only the review workflow would leave
the same class of hole open elsewhere.

### External branding removed

The merge step appended `Co-authored-by: Dopey <hello@sshdopey.com>` to every
auto-merged commit. Removed. The bot co-author filtering it was wrapped around
is kept.

The `"Lernza"` prompt text the issue mentions is already gone — the system
prompt reads "ProofOfHeart open-source project". `ROADMAP.md` still references
both strings, but that file is a dated audit snapshot ("captures the state of
`main` at commit `a9d0c3e`"), so it is left as written rather than rewritten
after the fact.

## Already resolved

The issue also lists `pull_request_target` + `contents: write` granting fork
PRs write access. That was fixed by the two-workflow split in #570 (closing
#472) and is unchanged here — verified in this branch:

- `pr-collector.yml` — `pull_request`, `contents: read` + `pull-requests: read`,
  no secrets, checks out `base.sha` so fork code never executes.
- `pr-review.yml` — `pull_request_review`, `contents: read` +
  `pull-requests: write`, secrets gated behind a trusted-reviewer check and the
  `/run-auto-review` phrase.

`grep -rn "pull_request_target\|--admin\|contents: write" .github/` returns only
explanatory comments.

## Verification

- All four workflows parse as YAML; every `run:` block passes `bash -n`.
- The gate's `jq` logic was exercised against seven mocked rollups: all-green
  (with this workflow's own check in progress), a failing check run, a pending
  check, a null rollup, a conflicting PR, a failing legacy `StatusContext`, and
  a rollup containing only this workflow's own check. Only the first yields
  `green=true`; the last two confirm the failure path and the self-deadlock
  guard.
- The merge-body pipeline was tested under `set -euo pipefail` with a normal
  body, a body that is entirely bot co-author trailers, and an empty body —
  none trip `pipefail`.

## Note for maintainers

Removing `--admin` only helps if branch protection on `main` actually requires
the CI checks. Worth confirming `fmt`, `clippy`, and `test` are configured as
required status checks — otherwise GitHub will still permit a merge over a red
build, and the new gate in this workflow would be the only thing stopping it.


davidmaronio (approved):
rebased and green, and the scope is exactly what was reviewed: --admin removed, fail-closed ci gate, sha-pinned actions. merging, thanks for the quick turnaround.

PR #700, opened 2026-07-30
fix: move require_not_paused before require_auth in accept_campaign_transfer (#453)

victor-134:
## Summary

Fixes #453.

**Problem:** In `accept_campaign_transfer`, `pending.require_auth()` was called **before** `require_not_paused(env)?`. This meant users would sign with Freighter only to get a `ContractPaused` error, wasting their signature.

**Fix:** Move `require_not_paused(env)?` to run **before** `pending.require_auth()`, so the paused check happens first and the user is never prompted to sign if the contract is paused.

## Changes

- `src/campaigns/transfer.rs`: Reordered checks in `accept_campaign_transfer` — paused check now precedes auth.


davidmaronio (changes requested):
the core change is right: hoisting `require_not_paused` to the top of `accept_campaign_transfer` (src/campaigns/transfer.rs:10) before the pending lookup and `require_auth` is exactly what #453 asked for, and putting it before the pending resolution is arguably the cleanest ordering of the open PRs.

a few things before this can move:

1. this is the same line move as #713 (issue #437), which is currently green and likely to merge first. once it does, this PR becomes redundant; please coordinate so we do not merge the same fix twice. if #713 lands, closing this one and confirming #453 on main is fine.

2. CI is red on fmt, clippy and test. the branch is based on an old main that still contained the broken placeholder files (src/campaigns.rs etc), so a rebase onto current main should fix compilation; please rerun `cargo fmt` and `cargo clippy` locally after.

3. the test_campaign_update.rs payload changes duplicate a fix already on main and should drop out after the rebase.

the merge state is dirty, so this needs conflicts resolved and CI green after a rebase on main, or closure if #713 merges first.

PR #583, opened 2026-07-03
Codex/502 add llvm cov coverage

Killerjunior:
## 📌 Description
Adds CI coverage reporting with `cargo-llvm-cov`, uploads LCOV results to Codecov, and enforces an 80% line coverage threshold gate.

## 🔗 Related Issues
Closes #502

## 🧪 Changes Made
- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation update

## ✅ Checklist
- [ ] Code compiles successfully
- [ ] Tests added/updated and passing
- [ ] Linting passes (no warnings/errors)
- [ ] Documentation updated (if required)
- [x] No breaking changes (or clearly documented)

## ⚠️ Breaking Changes
None.

## 📸 Screenshots (if applicable)
Not applicable.

## 🧩 Additional Notes
This adds a new `coverage` job to CI that installs `cargo-llvm-cov`, runs coverage with the `testutils` feature, generates `lcov.info`, fails below 80% line coverage, and uploads the report using `codecov/codecov-action`.

PR #704, opened 2026-07-30
feat: add cargo-llvm-cov coverage reporting with Codecov upload and 80% threshold gate

Seunfunmi-319509:
## Summary

Adds `cargo-llvm-cov` to the CI pipeline for automated code coverage reporting, uploading results to Codecov, and enforcing an 80% line coverage threshold gate.

Closes #502

### CI workflow (`.github/workflows/ci.yml`)
- **New `coverage` job** that runs alongside `fmt`, `clippy`, and `test`
- Installs `cargo-llvm-cov` via `cargo install cargo-llvm-cov --locked`
- Generates LCOV-formatted coverage report with `cargo llvm-cov --lcov`
- **80% threshold gate**: `--fail-under-lines 80` fails the pipeline if line coverage drops below 80%
- Uploads the LCOV report to [Codecov](https://about.codecov.io) using `codecov/codecov-action@v4`
- Uses `fail_ci_if_error: true` to fail the build if the upload itself fails

### Test fix (`src/tests/test_campaign_update.rs`)
- Fixed two tests (`test_update_campaign_emits_title_and_description`, `test_update_campaign_event_tracks_latest_description`) that were decoding the `campaign_metadata_updated` event payload as a 2-tuple `(String, String)` instead of the correct 4-tuple `(old_title, old_description, new_title, new_description)` — these were causing `HostError: Error(Object, UnexpectedSize)` panics

## Why this matters

- **Coverage visibility**: Without coverage tracking, critical paths (vesting reserve, token migration, personal cap, etc.) may have zero coverage
- **Quality gate**: The 80% threshold prevents coverage regressions from landing silently
- **Historical context**: Coverage tracking was already called out in `ROADMAP.md` as a priority — this implements that roadmap item

## Setup required

After merging, the repository admin must:
1. Add a `CODECOV_TOKEN` secret to the repository (from [codecov.io](https://codecov.io))
2. Enable the Codecov GitHub App integration


davidmaronio (changes requested):
nice setup overall: a dedicated coverage job with cargo-llvm-cov, an enforced `--fail-under-lines 80` gate independent of the Codecov upload, and `fail_ci_if_error: false` so a missing CODECOV_TOKEN cannot brick CI. that is a sensible split of enforcement vs reporting.

a few asks:

1. .github/workflows/ci.yml:41-44 - the new job uses `actions/checkout@v4` and `codecov/codecov-action@v4` by tag while #712 is pinning every action in this repo to a commit SHA. please pin these two the same way so we do not immediately reintroduce unpinned actions.

2. .github/workflows/ci.yml:47 - `cargo install cargo-llvm-cov --locked` compiles the tool from source on every run, which adds several minutes. `taiki-e/install-action@<sha>` with `tool: cargo-llvm-cov` (or a cargo cache) keeps the job fast.

3. worth double checking the current line coverage locally before merging; if main is below 80 the gate turns every unrelated PR red on day one. if it is close, consider starting at the measured floor and ratcheting up.

4. the src/tests/test_campaign_update.rs payload changes duplicate a fix already on main and should fall out of the diff on rebase.

also note #583 (Killerjunior) is the same llvm-cov idea; this version has the threshold gate and safer codecov settings, so it looks like the stronger candidate, but maintainers should explicitly pick one. this branch is in conflict with main, so please resolve conflicts and rebase, then address the pinning.

PR #569, opened 2026-06-30
fix(contract): enforce maximum campaign deadline extension

MerlinTheWhiz:
## 📌 Description

Adds `CAMPAIGN_EXTENSION_MAX_DAYS` — a non-configurable absolute cap of 365 days on total campaign duration (from start time to extended deadline) in `extend_campaign_deadline`. This acts as a defense-in-depth safety net, preventing the deadline from being pushed beyond 365 days even if admin-configured category duration caps are set to the maximum.

Previously, the category duration cap (default 365 days, configurable by admin up to 365) was the only guard on total campaign duration during extension. The new constant provides a hard-coded ceiling that cannot be overridden, isolating the extension logic from admin governance of category caps and preventing funds from being locked indefinitely.

## 🔗 Related Issues

Closes #558 
Closes #547 

## 🧪 Changes Made

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation update

## ✅ Checklist

- [x] Code compiles successfully
- [x] Tests added/updated and passing
- [x] Linting passes (no warnings/errors)
- [ ] Documentation updated (if required)
- [x] No breaking changes (or clearly documented)

## ⚠️ Breaking Changes

None. The new check is strictly additive — it only rejects extensions that were already rejected by the existing category cap check (both use 365 days as the default cap). All 274 existing tests pass unchanged.

## 📸 Screenshots (if applicable)

N/A

## 🧩 Additional Notes

The constant is placed alongside the existing validation constants in `src/lib.rs` and is referenced directly in `src/campaigns/update.rs` without passing through admin-configurable storage, ensuring the 365-day ceiling is always enforced.

PR #706, opened 2026-07-30
Fix #463: extend_campaign_deadline validates new deadline against category duration cap

Seunfunmi-319509:
## Summary

This PR adds a regression test for issue #463, which reported that `extend_campaign_deadline` did not validate the new deadline against the category duration cap when the cap was set AFTER the campaign was created.

## Current State

The fix is **already implemented** in the codebase at `src/campaigns/update.rs:119-130`. The `extend_campaign_deadline` function fetches the category duration cap at extension time (not at creation time) and validates the total campaign duration against it:

```rust
let category_cap = get_category_duration_cap(env, campaign.category)
    .unwrap_or(crate::CAMPAIGN_DURATION_MAX_DAYS);

let total_duration_seconds = new_deadline
    .checked_sub(start_time)
    .ok_or(Error::Overflow)?;
let total_duration_days = total_duration_seconds / crate::SECONDS_PER_DAY;

if total_duration_days > category_cap {
    return Err(Error::InvalidDuration);
}
```

This ensures that even if a campaign was created before a category duration cap was set, extending its deadline cannot exceed the cap.

## Changes

- Added regression test `test_extend_deadline_cap_set_after_campaign_creation_rejected` in `src/tests/test_campaigns.rs` that verifies:
  1. A campaign created with 30 days duration (no cap set)
  2. Admin sets category cap to 30 days after campaign creation
  3. Attempting to extend by 10 days (total 40) is correctly rejected with `Error::InvalidDuration`
  4. Extending by 0 days is rejected because `additional_days` must be > 0

## Testing

All 10 deadline extension tests pass, including the new regression test.

closes #463


davidmaronio (changes requested):
good instinct on #463: the cap-set-after-creation scenario is exactly the kind of edge that deserves a regression test, and the new test in src/tests/test_campaigns.rs reads clearly.

but a few things need sorting out:

1. scope - this PR contains no change to `extend_campaign_deadline` itself; the only src change is the new test. since CI is green, main apparently already reads the category cap dynamically at extension time, which means #463 is already fixed and this is a regression test, not a fix. please retitle/describe it that way, or if you believe a code change is still needed, include it.

2. src/tests/test_campaigns.rs:1652-1658 - step 4 contradicts itself: the comment says "extending by 0 days (total 30) should succeed - exactly at cap" and then the assertion expects `Error::ExtensionTooLong`. pick one story; if 0-day extensions are invalid by design, drop the "should succeed" comment and keep the assertion.

3. the whole coverage job in .github/workflows/ci.yml is copied from your other PR #704 and has nothing to do with #463. stacked branches make review and merge ordering painful; please rebase this branch off main directly so the diff is just the test.

4. the test_campaign_update.rs payload changes also duplicate a fix already on main.

heads up that #569 covers the related absolute extension cap, so coordinate if both land. this branch is in conflict with main, so please resolve conflicts, drop the unrelated CI job, and fix the step 4 comment/assertion mismatch.

Seunfunmi-319509:
@davidmaronio, kindly review

PR #712, opened 2026-07-30
Fix #446: Remove --admin flag, require CI green, pin action SHAs, remove external branding

pixels26:
## Summary

The auto-review CI workflow (`pr-review.yml`) had several security issues that bypassed branch protection and used external branding.

## Changes

1. **Remove `--admin` flag** from merge command — was bypassing branch protection and required status checks
2. **Require CI green** before auto-merge — new step checks combined commit status via `/repos/{repo}/commits/{sha}/status` API
3. **Pin all action versions to commit SHAs** — supply-chain hardening for `dawidd6/action-download-artifact`, `actions/checkout`, `actions/upload-artifact` across all workflow files
4. **Remove external branding** — stripped `Co-authored-by: Dopey <hello@sshdopey.com>` from merge commit body

Close #446

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
   fmt:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
       - uses: dtolnay/rust-toolchain@stable
         with:
           components: rustfmt
@@ -22,7 +22,7 @@ jobs:
   clippy:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
       - uses: dtolnay/rust-toolchain@stable
         with:
           components: clippy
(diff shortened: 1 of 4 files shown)


davidmaronio (changes requested):
thanks for this, the SHA pinning across ci.yml, fuzz.yml, pr-collector.yml and pr-review.yml is exactly what #446 asked for, and dropping `--admin` from the auto-merge so branch protection actually applies is a real security win. the co-author body rewrite removal also simplifies things nicely.

a few asks though:

1. .github/workflows/pr-review.yml:109 - the new CI gate uses `gh api "repos/$REPO/commits/$HEAD_SHA/status"`, but that is the legacy commit *status* API. GitHub Actions jobs report as *check runs*, not statuses, so on a repo where CI is only Actions this endpoint returns `state: "pending"` with zero contexts forever. that means `ci_green` will never be true and auto-merge silently stops working. use check runs instead, e.g. `gh api "repos/$REPO/commits/$HEAD_SHA/check-runs" --jq '[.check_runs[].conclusion] | all(. == "success" or . == "neutral" or . == "skipped")'` or simply `gh pr checks "$PR_NUM" --json state`.

2. .github/workflows/fuzz.yml:51 and pr-review.yml:86 - the pin comments say `# v4` and `# v6` while checkout gets a precise `# v4.4.0`. worth making them all exact (`# v4.6.x`, `# v6.x.y`) so future bump PRs can be sanity checked against the comment.

3. the src/admin.rs, src/campaigns.rs, src/lib.rs deletions and the test_campaign_update.rs payload fixes look like cleanup that already landed on main (that is why the branch shows behind). after a rebase they should disappear from this diff, keeping the PR purely CI-focused.

once the check-runs fix is in, please rebase on main so the unrelated cleanup drops out of the diff, then this is good to go.

PR #717, opened 2026-07-30
refactor(queries): extract shared bucket-pagination helper for get_ca…

Nife-tanny:
## Summary

Extract the identical bucket-traversal algorithm previously duplicated in `get_campaigns_by_category` and `get_creator_campaigns` into a single private helper, parameterised by total count, bucket size, and a bucket-getter closure.

**Update after review:** the extraction itself has since landed on main via #734 (as `get_campaigns_from_buckets`, covering both callers). This PR's incremental contribution is now:

- the algorithm documentation for the shared helper, and
- a code comment pinning the intentional `if let Some` sparse-bucket behaviour.

## Algorithm (unchanged)

1. Jump to the bucket containing the page offset.
2. Walk entries within that bucket starting at the requested position.
3. Collect up to `limit` campaigns (capped at `LIST_MAX_LIMIT`).
4. When the bucket is exhausted, advance position past the bucket boundary and repeat from step 1 with the next bucket.

## Behaviour notes (reviewer follow-ups)

- **Intentional:** the category path uses `if let Some(campaign_id) = bucket.get(idx_in_bucket)` (the safer form the creator path already used) rather than `unwrap()`. A sparse bucket — one whose entry is absent — now **skips** the slot instead of panicking. This is deliberate so both paths behave identically.
- `get_campaigns_by_category` now reads the category total before the `limit == 0` early return — one extra storage read on a degenerate call, understood and accepted.

## Validation

- Public function signatures unchanged — backwards-compatible.
- Only `src/queries.rs` modified.
- `cargo fmt --check` and `cargo clippy --all-targets --features testutils -- -D warnings` pass cleanly.
- Full suite: 421 tests pass, including the rebased 4-tuple `test_campaign_update` payload assertions.

Closes #663


davidmaronio (approved):
both notes addressed, the doc comment on the helper and the if-let rationale read well. extraction is still faithful after the rebase. merging.

PR #723, opened 2026-07-31
Fix #451: creator absorbs remaining pool as final revenue claimant

meem08:
## Summary

Fixes **#451** ([Security]) — `claim_creator_revenue` can under-pay the creator (the final revenue claimant) due to cumulative integer-division precision loss.

The creator's share is computed with truncating integer division (`floor(total_pool * creator_share_bps / BPS)`), the same pattern the issue flags for contributors. Across deposit rounds, cumulative rounding can drain the pool before the creator receives their full allocation, leaving revenue dust permanently stuck in the contract — the creator's last claim silently returns `0`/less than their entitlement.

### Root cause

Each distribution truncates fractional stroops. The sum of every contributor's individually-truncated share plus the creator's individually-truncated share can fall short of the full pool. Once the pool stops growing, nobody can claim the leftover stroops.

### The fix

As proposed in the issue — **track the distributed amount as a running sum and give the final claimant `pool_remaining` instead of their calculated share**:

- Once every contributor has claimed at least once (`contributors_claimed >= total_contributors`), the creator — the last party that can still claim — receives whatever remains of the pool (`total_pool - contributor_distributed - creator_already_claimed`) instead of their individually-truncated share, so the full allocation is paid out exactly and no dust is left stuck.
- This reuses the running-sum counters introduced for #526 (`contributor_revenue_distributed`, `contributor_revenue_claimants`) — no new storage keys.
- All arithmetic uses `checked_add`/`checked_sub` → `Error::Overflow` on pathological pools, never a panic (#408 pattern).

### Caveat (documented in code)

The guard assumes contributors claim **before** the creator after each deposit round. If the creator claims creator-revenue before contributors have claimed their share of a fresh deposit, `pool_remaining` includes their still-unclaimed share. Contributors should claim promptly after each deposit to keep this invariant.

## Changes

| File | Change |
|------|--------|
| `src/revenue.rs` | `claim_creator_revenue`: final-claimant dust absorption (running-sum `pool_remaining`), with ordering caveat comment |
| `src/tests/test_regressions.rs` | Regression test `test_creator_final_claimant_absorbs_rounding_dust` |

## CI repair commits (this branch)

The branch was updated against upstream `main` and three follow-up commits were required to get all CI checks green — **upstream `main` itself is currently red** (all checks fail) because of commit `900bb53` (#618, campaign tag filtering), which accidentally added a duplicate `src/campaigns.rs` alongside the `src/campaigns/` directory (rustc E0761) plus duplicate `#[contractimpl]` blocks in `admin.rs`/`lib.rs` (E0252/E0428) and non-Rust garbage files (`src/clients.rs` containing TypeScript, `src/events.ts`, `src/types.ts`, `src/test.rs`).

| Commit | Why |
|--------|-----|
| `27af3a0` Remove stale `src/campaigns.rs` duplicate module | The file (from #618) duplicated the `src/campaigns/` module → E0761 broke fmt/clippy/test |
| `5512fc9` Revert broken #618 commit `900bb53` | Reverts the duplicate `#[contractimpl]` blocks and deletes the non-Rust garbage files; keeps the legit #699/#616 upstream changes that were also in the merge |
| `f1ff24a` Fix `campaign_metadata_updated` event shape in stale tests | Two pre-existing `test_campaign_update` tests still parsed the event payload as a 2-tuple; the code (#510/#602) publishes `(old_title, old_description, new_title, new_description)` |
| `016a437` Fix clippy `unit_cmp` lint | Pre-existing `assert_eq!(data, ())` in `test_withdrawals.rs` trips `clippy::unit_cmp` under `-D warnings` |

If maintainers prefer, the `5512fc9` revert of #618 could be split into its own PR, but keeping it here is what makes this branch mergeable (a PR against broken `main` cannot go green otherwise).

## Test plan

- `cargo fmt --all -- --check` ✅
- `cargo clippy --all-targets --features testutils -- -D warnings` ✅
- `cargo test --features testutils` — **401 passed, 0 failed** ✅
- `cargo build --target wasm32-unknown-unknown --release` ✅ (the test job's second step)
- New regression test `test_creator_final_claimant_absorbs_rounding_dust` passes and covers: contributor claims full share across two deposit rounds, then creator absorbs the remaining pool (7 stroops) instead of their truncated share, leaving the contract fully drained

Closes #451

### DIFF
--- a/src/revenue.rs
+++ b/src/revenue.rs
@@ -6,11 +6,12 @@ use crate::lifecycle::{
     token_client,
 };
 use crate::storage::{
-    bump_instance_ttl, get_contribution, get_contributor_count, get_contributor_revenue_claimants,
-    get_contributor_revenue_distributed, get_creator_revenue_claimed, get_revenue_claimed,
-    get_revenue_pool, get_token, set_contributor_revenue_claimants,
-    set_contributor_revenue_distributed, set_creator_revenue_claimed, set_revenue_claimed,
-    set_revenue_pool,
+    bump_instance_ttl, get_contribution, get_contributor_count,
+    get_contributor_revenue_distributed, get_contributor_revenue_round_claimants,
+    get_contributor_revenue_round_pool, get_creator_revenue_claimed, get_revenue_claimed,
+    get_revenue_pool, get_token, set_contributor_revenue_distributed,
+    set_contributor_revenue_round_claimants, set_contributor_revenue_round_pool,
+    set_creator_revenue_claimed, set_revenue_claimed, set_revenue_pool,
 };
 
 pub(crate) fn deposit_revenue(env: &Env, campaign_id: u32, amount: i128) -> Result<(), Error> {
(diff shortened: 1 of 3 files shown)


davidmaronio (changes requested):
there's a lot to like here: reusing the #526 running-sum counters instead of new storage keys, all arithmetic checked, CEI preserved, and `test_creator_final_claimant_absorbs_rounding_dust` is a genuinely good regression test (the 7-vs-6 stroop assertion plus the contract-balance-zero check pins the behavior down precisely). the PR description is also unusually clear about upstream state.

but the caveat you documented is not a caveat, it's a theft vector, and it has to be closed before this merges:

1. src/revenue.rs:200: the gate is `contributors_claimed >= total_contributors`, and `contributor_revenue_claimants` counts unique claimants over the campaign's lifetime, not per deposit round. so after round 1, once every contributor has claimed once, the condition is permanently true. on every later deposit the creator can call `claim_creator_revenue` before contributors claim their share of the fresh deposit, and `pool_remaining = total_pool - distributed - already_claimed` includes the contributors' unclaimed round-N entitlement. the creator sweeps it, and the contributors' subsequent claims underpay or fail. "contributors should claim promptly" isn't an invariant the contract enforces, so a malicious creator just deposits and immediately claims.
2. suggested fix: gate on amounts, not claimant counts. only absorb the remainder when the contributor side of the *current* pool is fully paid out, e.g. compute the contributor-side entitlement (`total_pool - creator_share_total`, which the #526 last-claimant logic already pays out exactly) and require `contributor_revenue_distributed >= total_pool - creator_share_total` before letting the creator take `pool_remaining`. otherwise fall back to the truncated `claimable`. that makes the dust absorption safe regardless of claim ordering.
3. test to add: creator deposits round 2 and claims immediately, before the contributor. assert the creator gets only their truncated share and the contributor can still claim their full round-2 entitlement afterwards. that's the exploit path, and it should be red on the current code.
4. src/campaigns.rs, src/clients.rs, src/events.ts, src/types.ts, src/test.rs, src/admin.rs, src/lib.rs: the #618 revert commits were the right call at the time, but main has since cleaned itself up, which is why the PR now shows conflicts. after a rebase these hunks should disappear and the diff should shrink to revenue.rs + test_regressions.rs (the test_campaign_update.rs and test_withdrawals.rs fixes may also already be on main, keep them only if still needed).

gate: fix the ordering hole with an amount-based gate, add the creator-claims-first test, then rebase onto current main (`git fetch upstream && git rebase upstream/main && git push --force-with-lease`) and get CI green. the mechanism is right, it just needs the gate tightened.

meem08:
@davidmaronio  kindly merge

PR #725, opened 2026-07-31
feat: add emergency pause multi-sig endpoint

bbkenny:
Resolves #561

This PR adds an `emergency_pause` endpoint allowing a designated list of multi-sig emergency signers to pause the contract instantly. Unpausing still requires full admin authority. 

It also removes stray struct implementations (e.g. `ProofOfHeartContract`) left over from previous refactors that caused CI failures in the previous PR.

**Changes:**
- Added `EmergencySigners` to `AdminKey` in `storage.rs`
- Added `add_emergency_signer` and `remove_emergency_signer` (admin only)
- Added `emergency_pause` (requires auth from any authorized emergency signer)
- Wrote integration tests verifying the pause behavior and authorization limits

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Added
+
+- Multi-sig emergency pause capability via `add_emergency_signer`, `remove_emergency_signer`, and `emergency_pause`. This allows any authorized emergency signer to pause the contract instantly, while unpausing still requires full admin authority (#561).
+
 ### Fixed
 
 - `cancel_campaign` now rejects with `GoalMetCancellationNotAllowed` when `amount_raised >= funding_goal` and funds have not yet been withdrawn, preventing rug-pull-adjacent behaviour where a creator could cancel after reaching the goal and force all contributors to self-serve refunds (#164).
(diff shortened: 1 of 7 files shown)


davidmaronio (changes requested):
this is a much more careful take on emergency pause than the last attempt, nice work. the data model is safe this time (a plain `Vec<Address>` under a new `AdminKey::EmergencySigners` instance key, no new `#[contracttype]` struct fields), unpause stays admin-only, and there are real authorization tests. a few asks before merge:

1. src/tests/test_admin.rs:858-871 (test_emergency_pause_success) - there's a big block of stream-of-consciousness comments left in the test ("wait, unpause checks...", "Actually, we just need to verify..."). please delete that and replace it with one line saying unpause remains admin-only.
2. src/admin.rs:emergency_pause - `signer.require_auth()` runs before the membership check, which means any random address can be made to burn an auth entry probing this endpoint. cheap fix: check `signers.contains(&signer)` first, then `require_auth`. not a security hole, just tighter.
3. src/admin.rs:add_emergency_signer - the signer list is unbounded. worth a small cap (say 10) so an admin mistake can't bloat instance storage; instance entries have a size budget.
4. the pr also deletes the orphaned `ProofOfHeartContract` blocks in lib.rs/admin.rs/campaigns.rs and reworks the event assertions in test_campaign_update.rs. your pr #724 makes the exact same cleanup, so whichever lands second will conflict. pick one pr to own that cleanup and rebase the other on top.

gate: the branch is DIRTY against main, so resolve conflicts and rebase; once the test comments are cleaned up and the overlap with #724 is sorted this looks mergeable.

PR #726, opened 2026-07-31
fix: explicit CEI documentation for claim_refund to close #546

bbkenny:
Closes #546

Summary of Changes
Added a comment to `claim_refund` in `src/contributions.rs` explicitly documenting that the token transfer follows the Checks-Effects-Interactions (CEI) pattern. `claim_refund` was already safe and implemented the CEI pattern correctly (i.e. updating state by removing contribution records before initiating the transfer), preventing a malicious token contract from double-claiming refunds. This PR serves to explicitly document this to close the lingering security issue.


davidmaronio (changes requested):
good instinct to close #546 with an explicit note rather than churning code that was already safe, and the comment you added at the transfer site in src/contributions.rs:293 is accurate and in the right place. issues:

1. CI is fully red (fmt, clippy, test). the test job fails with E0761 "file for module `campaigns` found at both src/campaigns.rs and src/campaigns/mod.rs" plus a pile of duplicate-definition errors from the orphaned `ProofOfHeartContract` blocks in src/lib.rs and src/admin.rs. this pr keeps those orphan blocks (it only adds trailing newlines to them), while your own #724/#725 delete them. this branch needs that same cleanup, or a rebase once whichever pr owns the cleanup lands.
2. src/tests/test_campaign_update.rs - the 2-tuple to 4-tuple event assertion change is copy-pasted into this pr too (same as #724 and #725). a docs-only pr shouldn't be touching test assertions; drop it here and let one pr carry it.
3. small wording nit on the comment: the soroban host doesn't allow classic reentrancy today, so maybe phrase it as "state is updated before the transfer (CEI) so a misbehaving token cannot observe stale contribution records or double-claim via any future reentrancy allowance". keeps the doc honest about the actual threat model. optional.

gate: branch is DIRTY and CI is red, so please resolve conflicts, rebase on main, and get the three CI jobs green (that mostly means sorting out the orphan-code cleanup overlap with your other prs).

PR #728, opened 2026-07-31
feat: add campaign comment system via off-chain attestation (#542)

bbkenny:
Closes #542

Summary of Changes
Added a campaign comment system using off-chain attestation to keep storage costs low.

- **`add_campaign_comment`**: Allows users to post a comment to an active campaign. The function takes an off-chain `comment_hash` (which could be a CID or text hash depending on the frontend) and emits an event `campaign_comment_added` without storing the full string on-chain.
- **`remove_campaign_comment`**: Gives moderation tools to campaign creators and admins. This flags a specific `comment_id` as removed using a new `CommentRemoved` data key and emits a `campaign_comment_removed` event, signaling the indexer or frontend to hide it.
- Added `src/campaigns/comments.rs` for the logic and updated `errors.rs` to track invalid or previously removed comment targets cleanly.


davidmaronio (changes requested):
the design direction here is right: hash-only comments emitted as events with a per-campaign counter and a tombstone key for moderation keeps on-chain storage tiny, and the creator-or-admin removal gate plus double-remove protection is sensible. but the branch does not compile, and a few things look like they were never run locally:

1. src/errors.rs:148-155 - the new variants `InvalidCommentId = 46` and `CommentAlreadyRemoved = 47` were pasted inside the `name()` match arm block of `impl Error`, not into the `Error` enum itself. this is a syntax error and is why CI is red. move the variants into the enum and add matching arms to `name()`.
2. src/tests/test_comments.rs:5 - the tests use `ProofOfHeartContractClient` and a `setup_env_with_active_campaign(100)` helper returning a 5-tuple. please double check those against src/tests/helpers.rs on this branch; other test files in the repo use a different setup shape, and since CI never got past the errors.rs syntax error these tests have never actually compiled or run. please run `cargo test --features testutils` locally and paste the output.
3. src/campaigns/comments.rs:add_campaign_comment - anyone can spam comments for free apart from tx fees, and each one bumps a persistent counter. that's fine, but consider requiring the campaign to be active (you only check existence) so closed or cancelled campaigns don't keep accepting comments.
4. same file, remove flow - removal only writes a tombstone and emits an event, which matches the off-chain attestation model, but the `CommentRemoved` tombstone grows one persistent entry per removal forever. acceptable, just worth a comment noting the trade-off.
5. this pr also carries the unrelated test_campaign_update.rs 2-tuple to 4-tuple event assertion change and the trailing-newline touches to the orphaned `ProofOfHeartContract` blocks, same as your #724/#725/#726. drop the duplicates and let one pr own that cleanup, then rebase the others.

gate: branch is DIRTY and CI is red, so fix the errors.rs compile break, verify the tests actually run, resolve conflicts, and rebase on main.

PR #729, opened 2026-07-31
fix(#461): add regression tests for verify_with_votes overflow handling

OpensrcLord:
## Summary

Fixes #461

The checked-arithmetic fix for `verify_with_votes` was already in place — both aggregations use `checked_add(...).ok_or(Error::Overflow)?`:

- `src/voting.rs` — `approve_votes + reject_votes` (vote-count total)
- `src/voting.rs` — `approve_weight + reject_weight` (token-weighted total)

This PR adds regression tests to lock in that behavior so aggregation returns `Error::Overflow` instead of silently wrapping on `u32`/`i128` overflow:

- `test_verify_with_votes_count_overflow_returns_overflow` — forces `approve_votes + reject_votes` past `u32::MAX`
- `test_verify_with_votes_weight_overflow_returns_overflow` — forces `approve_weight + reject_weight` past `i128::MAX`

## Test plan

- [ ] `cargo test --features testutils` (CI runs this)

### DIFF
--- a/src/tests/test_campaign_update.rs
+++ b/src/tests/test_campaign_update.rs
@@ -66,6 +66,8 @@ fn test_update_campaign_emits_title_and_description() {
     let payload: (String, String, String, String) =
         soroban_sdk::FromVal::from_val(&env, &last_event.2);
 
+    assert_eq!(payload.0, String::from_str(&env, "Original Title"));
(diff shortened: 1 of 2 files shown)


davidmaronio (approved):
the shared cleanup hunks dropped out after the rebase as expected, this is now purely the regression tests. merging.

PR #735, opened 2026-08-02
fix: return NotAuthorized for bookmark auth failures

TheFaith-Code:
## Summary

Fixes the bookmark authorization regression for saved-campaign removal by ensuring an auth mismatch returns `NotAuthorized` instead of bubbling up as a host panic.

## What changed

- Updated bookmark save/remove handling to map unauthorized `require_auth()` failures to the contract's `Error::NotAuthorized`.
- Added a regression test covering removal of a bookmark for a different user, confirming the expected authorization failure.
- Cleaned up the contract module wiring so the Rust contract builds correctly again.

## Why

Previously, bookmark removal could fail with a host auth panic when the supplied address did not have the required authorization. This made the behavior inconsistent with the contract’s intended error model and made the regression hard to test.

## Verification

- `cargo fmt --check`
- `cargo test --features testutils`

Closes #672


davidmaronio (changes requested):
thanks for digging into this, the writeup shows real investigation into the panic path and the regression test for removing someone else's bookmark is a scenario we genuinely didn't cover. a few things need to change before this can land though:

1. src/bookmarks.rs:20-35: the catch_unwind path only exists under cfg(feature = "testutils"), so production wasm still traps with a host auth error and never returns Error::NotAuthorized. that means issue #672 isn't actually fixed on-chain, and the test now asserts behavior the deployed contract doesn't have. please drop require_auth_or_not_authorized entirely and keep plain user.require_auth() in the contract.
2. src/tests/test_bookmarks.rs:123-125: you can write this regression test with no contract change. without mock_all_auths, client.try_remove_saved_campaign(&contributor2, &campaign_id) returns an invoke/host auth error you can assert on directly. that keeps test and production semantics identical.
3. src/bookmarks.rs:25-28: even if we kept the wrapper, catch_unwind maps every panic to NotAuthorized, so any future unrelated panic in that closure becomes a fake auth error in tests.
4. src/tests/test_bookmarks.rs:94-116: prefer setup_env + create_campaign over hand-building a Campaign struct with set_campaign, so the test doesn't silently drift when the struct gains fields.

happy to re-review once the contract-side wrapper is removed and the test asserts the host auth failure instead.

davidmaronio (approved):
confirmed the wrapper is fully removed, the diff is now test only and the env.auths() assertion is the right way to pin the auth requirement without fighting mock auth. the campaignnotbookmarked check is a nice bonus. merging.

PR #741, opened 2026-08-05
Feature/campaign stats clean

Emelie-Dev:
closes #538


davidmaronio (approved):
this addresses everything from the #585 review: the datakey variants actually exist now, the setters are wired into contribute(), the entrypoint is in lib.rs, and the tests cover empty state, top contributor flipping, and the no-regress case. two soft notes for a follow-up: restore the doc comments dropped from the existing contributionkey variants, and a one-liner that top_contributor can go stale after a refund would help readers. merging.

PR #743, opened 2026-08-05
fix: return per-id (verified_ids, failed_ids) from verify_campaigns batch (#442)

dedukpe:
## 📌 Description

Fixes #442. The batch admin verification entry point `verify_campaigns` previously captured only the **first** per-campaign error and kept going, then returned `Err(first_error)`. That collapsed a partial-success batch into a single opaque error: callers received a `campaigns_bulk_verified` event with only `(verified_count, total)` and had **no way to tell which specific campaigns failed** — or even whether anything was committed, since on-chain the `Err` reverted the entire batch.

`verify_campaigns` now returns the per-id outcome for every campaign it processed:

```rust
pub fn verify_campaigns(
    env: Env,
    campaign_ids: soroban_sdk::Vec<u32>,
) -> Result<(soroban_sdk::Vec<u32>, soroban_sdk::Vec<u32>), Error>
//                    ^ verified_ids      ^ failed_ids
```

- **Partial success is preserved and reported.** Successful verifications are committed even when other ids in the batch fail, so callers can distinguish "first failed, rest succeeded" from "all failed" and retry only the ids in `failed_ids`.
- **Per-campaign failures never abort the batch.** Unknown ids, already-verified campaigns, cancelled campaigns, etc. are collected in `failed_ids`; only hard errors (caller not admin, contract paused) still return `Err`.
- **Voting-state TTL extension is unchanged** — every processed id gets its TTL bumped, success or failure (matching `verify_campaign`).
- **Event payload upgraded:** `campaigns_bulk_verified` data changed from `(verified_count, total)` to `(verified_count, failed_ids)` so off-chain indexers can see exactly which campaigns failed without re-querying.

## 🔗 Related Issues

Closes #442

## 🧪 Changes Made

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation update

## ✅ Checklist

- [x] Tests added/updated and passing (415 passed, 0 failed)
- [x] Linting passes (`cargo fmt --check` + `cargo clippy --all-targets --features testutils -- -D warnings`)
- [x] WASM release build passes (`cargo build --target wasm32-unknown-unknown --release`)
- [x] Documentation updated (EVENT_PAYLOADS.md, ROADMAP.md, CHANGELOG.md)
- [x] Breaking changes clearly documented (see below)

## ⚠️ Breaking Changes

1. **Contract ABI** — `verify_campaigns` return type changed from `u32` (verified count) to `(Vec<u32>, Vec<u32>)` (`verified_ids`, `failed_ids`). Callers using the old `u32` signature must update.
2. **Event payload** — `campaigns_bulk_verified` data changed from `(verified_count: u32, total: u32)` to `(verified_count: u32, failed_ids: Vec<u32>)`. Indexers must update.
3. **Semantics** — a batch containing any failing id previously returned `Err` (and reverted the whole batch on-chain); it now returns `Ok` with the failed ids listed, and the successful verifications persist.

## 🧩 Additional Notes

- Tests updated in `src/tests/test_voting.rs` and `src/tests/test_lifecycle.rs`; added coverage for per-id partial-failure reporting, all-failed reporting, a cancelled campaign mid-batch, and the new event payload shape.
- Also resolves the P2 roadmap item "`verify_campaigns` (batch) silently skips failed entries past the first" in `ROADMAP.md`.


davidmaronio (approved):
this is the strongest version of the #442 fix so far. per id tuple return, partial success committed instead of reverted, event payload carries failed_ids, and the docs all updated to match. the cancelled-in-batch and all-failed tests cover the cases #606 missed. merging.

PR #751, opened 2026-08-24
Fix #438: zero effective_amount_raised on cancelled campaigns

meem08:
Fixes #438

## Summary

`cancel_campaign` (and `admin_cancel_campaign`, which shares the same bug) left `campaign.effective_amount_raised` at its pre-cancel value. Only individual `claim_refund` calls decremented it, so if contributors never claimed, the value stayed non-zero on a dead campaign — breaking indexer "live contributions" metrics for cancelled campaigns.

## Changes

- **`src/campaigns/cancel.rs`** — both `cancel_campaign` and `admin_cancel_campaign` now set `campaign.effective_amount_raised = 0` at cancellation, before persisting the campaign.
- **`src/contributions.rs`** — `claim_refund` now skips the `effective_amount_raised` deduction when the campaign is cancelled. Since the gauge is already zeroed at cancel time, subtracting from it would drive it negative. Failed-due-to-goal refunds (non-cancelled) keep the existing `checked_sub` behaviour unchanged.
- **Tests** — regression tests for both cancel paths asserting the gauge is zeroed on cancel and stays at zero after a refund:
  - `test_cancel_campaign_zeroes_effective_amount_raised` (`src/tests/test_campaigns.rs`)
  - `test_admin_cancel_campaign_zeroes_effective_amount_raised` (`src/tests/test_admin_cancel.rs`)
- **`CHANGELOG.md`** — entry added under `[Unreleased] → Fixed`.

## Verification

- `cargo fmt --all -- --check` — clean
- `cargo clippy --all-targets --features testutils -- -D warnings` — clean
- `cargo test --features testutils` — 397 passed, 2 failed. The 2 failures (`test_update_campaign_emits_title_and_description`, `test_update_campaign_event_tracks_latest_description`) are **pre-existing on `main`** (verified by stashing these changes and re-running against a clean checkout) — they fail on a `HostError: Error(Object, UnexpectedSize)` decoding the `campaign_updated` event payload and are unrelated to this PR.


meem08:
@ Kindly merge

